PR #17, opened 2023-07-13
Fix to spack_packages version option

### LINKED ISSUE #6: Allow to build a package with a specific version/commit of ACCESS-NRI/spack_packages
For reproducibility, it might be necessary to build a package using an older version of ACCESS-NRI/spack_packages.

CodeGat:
#6  was re-opened to accommodate some more required features (see https://github.com/ACCESS-NRI/build-ci/issues/6#issuecomment-1633570952) - namely, the ability to pick a spack_packages version from the model version of the workflow. 

This would mean that the built package would be of the form `build-<model>-<compiler><compiler version>-<spack_packages version>:latest`. For example: `built-oasis3-mct-intel2021.1.2-main`. 

Should close #6

### DIFF
--- a/.github/workflows/build-and-push-image-build.yml
+++ b/.github/workflows/build-and-push-image-build.yml
@@ -29,7 +29,7 @@ jobs:
     uses: access-nri/workflows/.github/workflows/build-and-push-image.yml@main
     with:
       container-registry: ghcr.io
-      container-name: access-nri/build-${{ matrix.package }}-${{ matrix.compiler.name }}${{ matrix.compiler.version }}
+      container-name: access-nri/build-${{ matrix.package }}-${{ matrix.compiler.name }}${{ matrix.compiler.version }}-${{ github.event.inputs.spack-packages-version }}
       dockerfile-directory: containers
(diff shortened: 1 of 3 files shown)


CodeGat:
If we are specifying spack_packages version in the built artifact, the artefact will look like `built-oasis3-mct-intel2021.1.2-main` (noting the `main` from the version of spack_packages). Is this okay?

aidanheerdegen:
> If we are specifying spack_packages version in the built artifact, the artefact will look like `built-oasis3-mct-intel2021.1.2-main` (noting the `main` from the version of spack_packages). Is this okay?

I think so.

Am I right, by default if `spack_packages` is updated and the CI is run the image will be built and overwrite the existing `main` image, e.g. `access-nri/built-oasis3-mct-intel2021.1.2-main`? Similarly a tagged build would also overwrite an existing package if the tag was updated?

CodeGat:
It seems that the older versions are still around @aidanheerdegen ,  but the latest one is tagged `latest` dynamically

aidanheerdegen (approved):
Looks good I think.

PR #18, opened 2023-07-18
13 add cice5

Quartzer2:
Although issues with dependencies/variants going to slack 0.20.1, this is just adding CICE5 to the base image.


CodeGat (approved):
Looks good to me :D

PR #19, opened 2023-07-18
Notification of spack_packages update for model image rebuild

### LINKED ISSUE #4: Trigger image rebuilds on ACCESS-NRI/spack_packages updates
Thanks @harshula for the suggestion.

We want to trigger our build images to rebuild whenever ACCESS-NRI/spack_packages has a relevant package file update. For example, if libaccessom2.py is updated we want to rebuild all libaccessom2 build images.

For a first pass implementation just a blanket rebuild on any push to spack_packages would be fine. 

Ideally we would check what package files are updated and only trigger the relevant image rebuild. Will need to handle dependencies (e.g. datetime-fortran, etc) somehow in this as well.

See https://github.com/orgs/community/discussions/26323 for an example of how to trigger workflows from an external repository.

CodeGat:
This PR handles the dispatched event from `ACCESS-NRI/spack_packages` that notifies `build-ci` that it has had some kind of update pushed to it and needs specific images to be rebuilt. 

Related PR from spack_packages: ACCESS-NRI/spack_packages#26

Should close #4 !


.github/workflows/notified-of-spack-packages-update.yml
```diff
@@ -0,0 +1,66 @@
...
+      packages: ${{ steps.generate.outputs.packages }}
+      compilers: ${{ steps.generate.outputs.compilers }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with: 
+          python-version: '3.10'
+      - name: Generate model matrix
+        id: generate
+        run: python ./util/matrix-generator.py ${{ github.event.inputs.packages }}
```
Quartzer2, line 36: In order to verify this python script, should we have a separate test for it?

.github/workflows/notified-of-spack-packages-update.yml
```diff
@@ -0,0 +1,66 @@
...
+          dockerfile-directory: containers
+          dockerfile-name: Dockerfile.build
+          build-args: |
+            # TODO: Probably shouldn't hard code base image path
+            "BASE_IMAGE=ghcr.io/access-nri/base-spack-${{ github.event.inputs.spack-packages-version }}:latest"
+            "PACKAGE=${{ matrix.package }}"
+            "COMPILER_NAME=${{ matrix.compiler.name}}"
+            "COMPILER_PACKAGE=${{ matrix.compiler.package}}"
+            "COMPILER_VERSION=${{ matrix.compiler.version}}"
+    permissions:
```
Quartzer2, line 62: We are interested in why permissions need to be explicitly defined. Just interested!

util/full-dependency-mapping.txt
```diff
@@ -0,0 +1,9 @@
+datetime-fortran libaccessom2 cice5 mom5 access-om2
+oasis3-mct libaccessom2 cice5 mom5 access-om2
+libaccessom2 cice5 mom5 access-om2
+cice5 access-om2
+mom5 access-om2
+access-om2 
+access-om3 
+fcm
+gcom
```
Quartzer2, line 9: How are these updated? Again just for interest!

util/supported-dependency-mapping.txt
```diff
@@ -0,0 +1,6 @@
+datetime-fortran libaccessom2 cice5 mom5
+oasis3-mct libaccessom2 cice5 mom5
+libaccessom2 cice5 mom5
+cice5
+mom5
+access-om3
```
Quartzer2, line 6: Is this generated from the full dependency mapping? Do we need instructions on what goes in here and how it is updated?

Quartzer2 (approved):
Again looks good. Just some questions and we are having fun going through and learning the review procedure :)

.github/workflows/notified-of-spack-packages-update.yml
CodeGat: I actually don't know :D It was just how the original was done, so I copied it

util/full-dependency-mapping.txt
CodeGat: For now, they aren't updated. I did the dependent graph manually, through investigating the spack environment. It'd have to be updated manually if the dependents change. Issue raised here: https://github.com/ACCESS-NRI/build-ci/issues/21

util/supported-dependency-mapping.txt
CodeGat: Yes, it is generated from the full dependency mapping. Essentially it was just removing the models that we don't support at the moment (access-om2 among others). I will add a readme to the util folder, though,

.github/workflows/notified-of-spack-packages-update.yml
CodeGat: Yes, we should. I'll add one.

Quartzer2 (approved):

PR #25, opened 2023-07-19
Notified Workflow Fix

### LINKED ISSUE #24: Fixes for workflow kicked off by spack_packages
Some small issues have cropped up. 
- trying to call matrix_generator.py instead of matrix_generation.py
- trying to use the workflow in the repository rather than the one managed by the org (requires checkout to use the ones in the repo).

CodeGat:
Now that the entire workflow is testable, I've encountered a few bugs:
- called python script incorrectly (wrong name)
- called a dispatch-only workflow (now made callable)

Should close #24 !

### DIFF
--- a/.github/workflows/build-and-push-image-base-spack.yml
+++ b/.github/workflows/build-and-push-image-base-spack.yml
@@ -8,6 +8,13 @@ on:
         type: string
         default: "main"
         description: "Either a git tag or branch name for the ACCESS-NRI/spack_packages repository, which defaults to the main branch"
+  workflow_call:
+    inputs:
+      spack-packages-version:
+        required: true
+        type: string
+        default: "main"
+        description: "Either a git tag or branch name for the ACCESS-NRI/spack_packages repository, which defaults to the main branch"
 
 jobs:
   build-and-push-image:
--- a/.github/workflows/notified-of-spack-packages-update.yml
+++ b/.github/workflows/notified-of-spack-packages-update.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name : Update Spack Base Image
-        uses: ./.github/workflows/build-and-push-image-base-spack.yml@main
+        uses: access-nri/workflows/.github/workflows/build-and-push-image-base-spack.yml@main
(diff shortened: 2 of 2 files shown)


.github/workflows/build-and-push-image-base-spack.yml
```diff
@@ -8,6 +8,13 @@ on:
         type: string
         default: "main"
         description: "Either a git tag or branch name for the ACCESS-NRI/spack_packages repository, which defaults to the main branch"
+  workflow_call:
```
jo-basevi, line 11: Just out of interest, whats the difference between workflow_call vs workflow_dispatch?

jo-basevi (approved):
It all looks good to me

.github/workflows/notified-of-spack-packages-update.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           python-version: '3.10'
       - name: Generate model matrix
         id: generate
-        run: python ./util/matrix_generator.py ${{ github.event.inputs.packages }}
+        run: python ./util/matrix_generation.py ${{ github.event.inputs.packages }}
```
Quartzer2, line 36: Ha... we should have picked that up yesterday!

Quartzer2 (approved):
All good from me as well.

.github/workflows/build-and-push-image-base-spack.yml
CodeGat: Workflow_dispatch means that this workflow can be called via the GitHub CLI, the web interface, or the GitHub API
Workflow_call means that this workflow can be called via other workflows (or it can be turned into an action and used with the `uses: actions/some_action` keyword.

.github/workflows/notified-of-spack-packages-update.yml
CodeGat: :D I should have picked it up initially too!

.github/workflows/build-and-push-image-base-spack.yml
jo-basevi: Ok, thank you for that explanation!

PR #27, opened 2023-07-20
Added appropriate permissions for base-spack workflow call

CodeGat:
The build was failing with
> he workflow is not valid. .github/workflows/notified-of-spack-packages-update.yml (Line: 16, Col: 3): Error calling workflow 'access-nri/workflows/.github/workflows/build-and-push-image-base-spack.yml@main'. The nested job 'build-and-push-image' is requesting 'packages: write', but is only allowed 'packages: read'.

Granting explicit permissions should fix this.


aidanheerdegen:
`No description provided` 👀

Quartzer2 (approved):

PR #29, opened 2023-07-21
Merging Image-Building workflows into one file

### LINKED ISSUE #16: Combine base-spack and build image jobs in single workflow
No reason to have these in separate workflows - no idea why I made them so :sweat_smile:.

Combine jobs from [base-spack](https://github.com/ACCESS-NRI/build-ci/blob/main/.github/workflows/build-and-push-image-base-spack.yml) and [build](https://github.com/ACCESS-NRI/build-ci/blob/main/.github/workflows/build-and-push-image-build.yml) workflows.

[build](https://github.com/ACCESS-NRI/build-ci/blob/main/.github/workflows/build-and-push-image-build.yml) job should wait for base-spack finish before executing (`needs: build-and-push-image-base-spack`).

CodeGat:
Varvara mentioned that `build-and-push-image-base-spack.yml` and `build-and-push-image-build.yml` can be merged into a single file. Since there are a few things that call only the base-spack part and others that only build the image, we can create an input that makes part of it optional. 

Should close #16 !


.github/workflows/build-and-push-image-build.yml
```diff
@@ -3,14 +3,41 @@ name: Build and push access-nri/build images
...
         required: true
         type: string
         default: "main"
         description: "Either a git tag or branch name for the ACCESS-NRI/spack_packages repository, which defaults to the main branch"
 
 jobs:
-  build-and-push-image:
+  base-spack-build-and-push-image:
+    if: inputs.spack-packages-build-required
+    uses: access-nri/workflows/.github/workflows/build-and-push-image.yml@main
```
aidanheerdegen, line 20:
```suggestion
    uses: access-nri/build-ci/.github/workflows/build-and-push-image.yml@main
```
This really confused me, but I think the `workflows` repo was renamed `build-ci`, so it resolves to `build-ci` but should probably be changed for clarity (assuming I am correct)

aidanheerdegen: I can't suggest a change on unchanged text, but I'd make the same change at line 30 too.

.github/workflows/build-and-push-image-build.yml
```diff
@@ -3,14 +3,41 @@ name: Build and push access-nri/build images
 on:
   workflow_dispatch:
     inputs:
+      spack-packages-build-required:
```
aidanheerdegen, line 6: When/where is this dispatched?

aidanheerdegen (approved):
Just some questions and maybe a cosmetic change for clarity.

.github/workflows/build-and-push-image-build.yml
CodeGat: this will be an option when running a workflow from github - it'll just be a timesaver in the case where a model is updated, and we want to use an existing tagged base-spack image so we can reuse rather than rebuild every time. 
Was that what you meant?

.github/workflows/build-and-push-image-build.yml
CodeGat: Woah, that's good to know :D I thought it was some special auto-created repo that houses all of access-nri's workflows!

.github/workflows/build-and-push-image-build.yml
aidanheerdegen: Up to now it was only dispatched manually, is that right?

PR #32, opened 2023-07-24
image build pipeline secrets fix

### LINKED ISSUE #30: build-ci CI pipeline failing due to secrets not being valid
Run in question: https://github.com/ACCESS-NRI/build-ci/actions/runs/5617533683 

Secrets seem to be invalid and lead to the entire build failing. I checked the secrets and they hadn't seemed to be modified. 
Seems to be an issue with how they are formatted, maybe? Might need @echus input on this one.

CodeGat:
After a cursory glance at the CI pipeline, I noticed that I hadn't put in the build secrets in as a part of the call to build-and-push-image workflow for the image build. This is why the build is failing. 

Should close #30 !


CodeGat:
This branch is branching off #31 rather than main - closing this one and opening a new one that is correct

PR #42, opened 2023-07-31
Create Multimodel Build CI Testing Image

### LINKED ISSUE #9: Software installation in current images can only be done easily using Spack
Currently the packages get installed in directories whose name includes hashes. That means that the only practical way to use them is through Spack, i.e., one needs to do `spack load ...` before trying to use the packages in a build.

It would be nice to support other ways of building the software, for example, by using environment modules. Spack does install the module files, but they can't be used because neither lmod nor tcl modules are installed in the image.

Another way would be to install the packages as part of a Spack environment. This would provide a "view"  where all the packages are easily accessible. Better yet, check the Dockerfiles that Spack creates to automatically install an environment. It does it in a nice way that makes it very easy to use the packages and, at the same time, minimizes considerably the size of the docker images.

CodeGat:
After the discussions in https://forum.access-hive.org.au/t/ci-cd-planning-discussion/1035/3 we note that a multimodel, `spack env`-based Dockerfile is the way to go for future CI stuff. 
This will lead to a fair bit of refactoring and removal of much of the util folder. 

Should close #9 !


.github/workflows/build-and-push-image-build.yml
```diff
@@ -37,17 +37,10 @@ jobs:
   model-build-and-push-image:
     needs:
       - base-spack-build-and-push-image
-    if: needs.base-spack-build-and-push-image.result == 'success' || needs.base-spack-build-and-push-image.result == 'skipped'
+    if: always()
```
aidanheerdegen, line 40: Do we need an `if` at all if we're always running it?

.github/workflows/build-and-push-image-build.yml
```diff
@@ -58,13 +51,12 @@ jobs:
     uses: access-nri/build-ci/.github/workflows/build-and-push-image.yml@main
     with:
       container-registry: ghcr.io
-      container-name: access-nri/build-${{ matrix.package }}-${{ matrix.compiler.name }}${{ matrix.compiler.version }}-${{ github.event.inputs.spack-packages-version }}
+      container-name: access-nri/build-${{ matrix.compiler.name }}${{ matrix.compiler.version }}-${{ github.event.inputs.spack-packages-version }}
```
aidanheerdegen, line 54: At the risk of adding more back into the container name, if we're going for monolithic "per-model" containers the model name should figure somewhere I think. Probably in the name, in which case does it also have to be an argument to the workflow under `jobs:`?

The ideal should be that we can specify a different model, e.g. access-om3, and all of this would work for that monolithic build  as well.

Don't rule out the possibility that I've misunderstood how this works and this is not required.

containers/Dockerfile.build
```diff
@@ -1,30 +1,24 @@
 ARG BASE_IMAGE
 FROM ${BASE_IMAGE}
 
-ARG PACKAGE
-# Need to specify compiler package name for install and module name for build separately
-# e.g. for the latest intel this would be:
-# COMPILER_NAME=intel
-# COMPILER_PACKAGE=intel-oneapi-compilers
-ARG COMPILER_NAME
+ARG PACKAGE_NAMES="mom5 cice5 oasis3-mct libaccessom2"
```
aidanheerdegen, line 4: So *if* what I was saying above about specifying the model above is correct, here some of the dependency mapping type stuff might still be useful. 
At the very least a `yaml` file something like
```yaml
access-om2: mom5 cice5 oasis3-mct libaccessom2
```

Could be defined and then used to populated the package names here? An alternative would be to have multiple `Dockerfile.build` files named for each model, e.g. `Dockerfile.access-om2.build`

.github/workflows/build-package.yml
```diff
@@ -38,18 +38,30 @@ jobs:
...
-      run: |
-        spack -d install --only package --no-checksum ${{ inputs.package-name }}@=$GITHUB_HEAD_REF%${{ inputs.compiler-name }}@${{ inputs.compiler-version }}
-      if: github.event_name == 'pull_request'
+    - name: Upload lockfiles on failure
+      uses: actions/upload-artifact@v3
+      if: failure()
+      with: 
+        name: failed-lockfile-output
+        path: /opt/spack/var/spack/environments/${{ inputs.package-name }}/${{ inputs.package-name }}.*.spack.lock
+
```
aidanheerdegen, line 64: We're uploading logs of failed builds for debugging and so they don't get deleted when workflow logs are cleaned?

.github/workflows/build-package.yml
```diff
@@ -38,18 +38,30 @@ jobs:
...
+    - name: Get correct ref
+      run: echo "GH_REF=${{ github.event_name == 'pull_request' }} && ${{ github.head_ref }} || ${{ github.ref_name }}" >> $GITHUB_ENV
+    
+    - name: Activate environment
+      run: spack env activate ${{ inputs.package-name }}
+    
     - name: Build package via spack
+      run: spack -d install --add --only package --fail-fast --no-checksum ${{ inputs.package-name }}@=$GH_REF%${{ inputs.compiler-name }}@${{ inputs.compiler-version }}
+    
+    - name: Generate lockfiles on failure
```
aidanheerdegen, line 50: We don't want to generate lockfile regardless of installation success or failure? I don't exactly know what I want to do with them though. :)

aidanheerdegen (approved):
Assuming what I said wasn't a misunderstanding I think it should be addressed before merging. 

Otherwise merge away!

.github/workflows/build-and-push-image-build.yml
aidanheerdegen: Answered my own question: this is required if a previous step might error and you still want this to run

see https://stackoverflow.com/a/58859404

CodeGat: Yeah this is for the case where we skip the above base-spack-build-and-push-image (because that base-spack image already exists). We still need it to have pushed the image before we do any of the other dependency image stuff, though.

.github/workflows/build-and-push-image-build.yml
CodeGat: No, you're correct there. We'll need to specify the overall model name (access-om2, access-om3) in the build. Will update now

.github/workflows/build-package.yml
CodeGat: Might be worth doing them on both conditions, actually. Will update the repo.

.github/workflows/build-package.yml
CodeGat: Not the logs, just the lockfiles. I think it'd be useful to see what the difference is between the lockfile that we have, and the lockfile that it regenerates to install it correctly, so we could update dependencies

containers/Dockerfile.build
CodeGat: Dependency mapping still has a home, it seems! I'll update the build.

.github/workflows/build-package.yml
aidanheerdegen: Agreed. I don't know if "logs" was a typo on my part or auto-correct, but I did understand they were lockfiles.

containers/Dockerfile.build
aidanheerdegen: > Dependency mapping still has a home, it seems! I'll update the build.

Or have multiple Dockerfiles. It's not a terrible option.

.github/workflows/build-and-push-image-build.yml
CodeGat: Updated!

containers/Dockerfile.build
CodeGat: Added multiple dockerfiles :)

.github/workflows/build-package.yml
CodeGat: Added on both conditions

containers/Dockerfile.access-om3
```diff
@@ -0,0 +1,24 @@
+ARG BASE_IMAGE
+FROM ${BASE_IMAGE}
+
+ARG PACKAGE_NAMES="access-om3"
```
aidanheerdegen, line 4: To be consistent with `access-om2` we should use the model components as defined in here

https://github.com/COSIMA/access-om3/tree/master

Namely: CDEPS, CICE6, CMEPS, MOM6, WW3

But we don't have package files for any of those so we can't put them in there. So `access-om3` is a decent place-holder as it just defines a few dependencies.

containers/compilers.yaml
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    name: "intel",
+    package: "intel-oneapi-compilers",
+    version: "2021.1.2",
+  },
+]
```
aidanheerdegen: I definitely want the next PR to add a compiler to this and check it works ;)

aidanheerdegen (approved):
Looks good, just a couple of vague comments.

aidanheerdegen:
Feel free to resolve those conversations that are blocking the merge (once you're read them of course ;) )

containers/compilers.yaml
CodeGat: Will give that a go!

containers/Dockerfile.access-om3
CodeGat: Noted, will keep in mind when we add package files

PR #46, opened 2023-08-02
9 env based single model dockerfile

### LINKED ISSUE #9: Software installation in current images can only be done easily using Spack

CodeGat:
Dependency images created will now be done inside an environment, and push the `spack.yaml` and `spack.lock` files as a release of build-ci. 

Should close #9 !

### DIFF
--- a/.github/workflows/build-and-push-image-build.yml
+++ b/.github/workflows/build-and-push-image-build.yml
@@ -55,36 +55,18 @@ jobs:
             version: "2021.1.2",
           },
         ]
-    runs-on: ubuntu-latest
-    steps:
-      - name: Build and push dependency image
-        id: dependency-image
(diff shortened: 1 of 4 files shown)


CodeGat:
Closing this as multimodel is the way to go

PR #47, opened 2023-08-03
Added metadata LABELs and optimised some operations

### LINKED ISSUE #45: Refactor names of dependency images to use tags instead of a long name
**Description**
Dependency image names are becoming quite long as there are more moving parts. Right now, an image name is `container-name: access-nri/build-${{ matrix.package }}-${{ matrix.compiler.name }}${{ matrix.compiler.version }}-${{ github.event.inputs.spack-packages-version }}` (for example, `build-mom5-intel2021.1.2-2023.07.25`, and longer for branch-based `spack_packages` versions).
Consider using multiple `docker tag`s for each image that specifies the compiler name and version, and spack_packages version, which would allow shorter names (like `build-mom5` with tags `compiler-intel@2021.1.2` and `spack_packages-2023.07.25`)

**Implementation considerations** 
* How would we search for an image that satisfies both a name and a set of tags?
* How would we search for an image that has the same name and set of tags, that is a later build? Possibly utilizing the automatic `latest` tag
* How would we search for an image that has the same name and set of tags, that is an earlier build? Can we search by date?

CodeGat:
A less feature rich version of the linked issue. Added metadata LABELs to the image for future leverage in the CI/CD pipeline. Optimized some RUN commands. 

Should close #45 !

### DIFF
--- a/containers/Dockerfile.build
+++ b/containers/Dockerfile.build
@@ -18,14 +18,13 @@ LABEL au.org.access-nri.ci.base-spack-iamge ${BASE_IMAGE}
 SHELL ["docker-shell"]
 
 # Install required compiler
-RUN spack -d install ${COMPILER_PACKAGE}@${COMPILER_VERSION}
-
 # Load required compiler and generate site-wide compilers.yaml
-RUN spack load ${COMPILER_PACKAGE}@${COMPILER_VERSION} && \
-  spack compiler find --scope site
-
 # Build and install package dependencies
-RUN spack -d install --only dependencies ${PACKAGE}%${COMPILER_NAME}@${COMPILER_VERSION}
+RUN spack -d install ${COMPILER_PACKAGE}@${COMPILER_VERSION} \
+ && spack load ${COMPILER_PACKAGE}@${COMPILER_VERSION} \
+ && spack compiler find --scope site \
+ && spack -d install --only dependencies ${PACKAGE}%${COMPILER_NAME}@${COMPILER_VERSION} \
+ && spack gc -y
 
 # Push any uncached binaries to buildcache
 RUN spack -d buildcache create -a -m s3_buildcache `spack find --json | jq --raw-output .[].name`
--- a/containers/Dockerfile.build
+++ b/containers/Dockerfile.build
@@ -10,6 +10,10 @@ ARG COMPILER_NAME
 ARG COMPILER_PACKAGE
 ARG COMPILER_VERSION
 
+LABEL au.org.access-nri.ci.compiler ${COMPILER_PACKAGE}@${COMPILER_VERSION}
+LABEL au.org.access-nri.ci.package ${PACKAGE}
+LABEL au.org.access-nri.ci.base-spack-iamge ${BASE_IMAGE}
+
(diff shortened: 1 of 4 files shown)


containers/Dockerfile.build
```diff
@@ -10,21 +10,21 @@ ARG COMPILER_NAME
...
-RUN spack load ${COMPILER_PACKAGE}@${COMPILER_VERSION} && \
-  spack compiler find --scope site
-
 # Build and install package dependencies
-RUN spack -d install --only dependencies ${PACKAGE}%${COMPILER_NAME}@${COMPILER_VERSION}
+RUN spack -d install ${COMPILER_PACKAGE}@${COMPILER_VERSION} \
+ && spack load ${COMPILER_PACKAGE}@${COMPILER_VERSION} \
+ && spack compiler find --scope site \
+ && spack -d install --only dependencies ${PACKAGE}%${COMPILER_NAME}@${COMPILER_VERSION} \
+ && spack gc -y
```
aidanheerdegen, line 27: Can you comment (in the PR) why it is a good idea to do all this in a single `RUN` command. Just for general information for the morbidly curious.

aidanheerdegen (approved):
Looks good. Merge away!

containers/Dockerfile.build
CodeGat: So, as far as I understand it:
* Every time you execute a `RUN` command, it creates a new 'layer' on top of the image. New layers have a small space overhead when they are created. 
* When stuff is added _or modified or deleted_ in a layer, this takes up space, as later layers are essentially a delta of what was, and what was executed in the `RUN` command. 
  * So, for example, if we do `RUN dnf install -y complex_package`, it will install the package and all the dependencies, and some of the cruft that comes with installing things.  
  * If we then do `RUN dnf clean all`, it will remove some of the cruft and remove unneeded packages. But, crucially, the final image could be _the same size or larger_ after this, as the delta of installing cruft and then removing it across `RUN` commands makes it larger.
* So, if we do all the `spack` commands in one layer, including `spack gc` and other cleanup (garbage collection), it will be smaller as all the commands are being executed in the same layer, so the delta will be free of any add-then-later-remove parts of the filesystem.

CodeGat:
Regarding the decision to remove some of the run commands, as far as I understand it:

* Every time you execute a RUN command, it creates a new 'layer' on top of the image. New layers have a small space overhead when they are created.
* When stuff is added or modified or deleted in a layer, this takes up space, as later layers are essentially a delta of what was, and what was executed in the RUN command.
    * So, for example, if we do RUN dnf install -y complex_package, it will install the package and all the dependencies, and some of the cruft that comes with installing things.
    * If we then do RUN dnf clean all, it will remove some of the cruft and remove unneeded packages. But, crucially, the final image could be the same size or larger after this, as the delta of installing cruft and then removing it across RUN commands makes it larger.
* So, if we do all the spack commands in one layer, including spack gc and other cleanup (garbage collection), it will be smaller as all the commands are being executed in the same layer, so the delta will be free of any add-then-later-remove parts of the filesystem.
* It also helps with caching! Less layers leads to better caching decisions (from Dockers perspective)

PR #53, opened 2023-08-16
Removed setting of spack PATH as it is already done later

### LINKED ISSUE #51: Setup envs properly in base-spack Dockerfile
Rather than just setting the ENV path, use the spack-mandated `$SPACK_ROOT/share/spack/setup-env.sh` script in `Dockerfile.base-spack` line 17

CodeGat:
Small optimization - the spack environment is already set up as an entrypoint to the Dockerfile.

Should close #51 !


harshula (approved):
* I removed the `ENV PATH="$SPACK_ROOT/bin:$PATH"` in my spack-dev Dockerfile and there was no problem.

PR #57, opened 2023-08-17
base-os: update the OS first and remove /var/cache/dnf

harshula:


CodeGat (approved):
Looks good to me! Hopefully it makes the image a little smaller. As an idea for a future PR - is there any way to decouple what we need from the group install of "Development Tools" for further optimizations?

harshula:
> is there any way to decouple what we need from the group install of "Development Tools" for further optimizations?

Working on it right now. Using the list from the spack-dev Dockerfile as a starting point.

PR #59, opened 2023-08-17
Cache using gha rather than local

CodeGat:
Cache using the beta `type=gha` rather than `type=local` for individual layer caching.


harshula (commented):
* gha is explained: https://docs.docker.com/build/cache/backends/gha/
* mode is explained here: https://docs.docker.com/build/cache/backends/#cache-mode

PR #60, opened 2023-08-18
Move boto3 install out of Dockerfile.base-os and into Dockerfile.base-spack

harshula:
The dev Dockerfile doesn't need boto3.


CodeGat (approved):
Makes sense to me! Approved :)

PR #61, opened 2023-08-18
base-spack: add improvements from the dev Dockerfile

harshula:
* Set default value of SPACK_PACKAGES_REPO_VERSION to main
* Change granularity to generic.
* Delete repos.yaml. Use the spack repo add command.

### DIFF
--- a/containers/Dockerfile.base-spack
+++ b/containers/Dockerfile.base-spack
@@ -6,7 +6,7 @@ ENV SPACK_ROOT=/opt/spack
 ENV SPACK_PACKAGES_REPO_ROOT=/opt/spack_packages
 ENV GNUPGHOME=$SPACK_ROOT/opt/spack/gpg
 
-ARG SPACK_PACKAGES_REPO_VERSION
+ARG SPACK_PACKAGES_REPO_VERSION=main
 ARG SPACK_REPO_VERSION=v0.20.1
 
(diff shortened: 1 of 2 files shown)


containers/Dockerfile.base-spack
```diff
@@ -15,10 +15,13 @@ LABEL au.org.access-nri.ci.spack-repo-version ${SPACK_REPO_VERSION}
 # Install spack
 RUN git clone -c feature.manyFiles=true https://github.com/spack/spack.git $SPACK_ROOT --branch ${SPACK_REPO_VERSION} --single-branch --depth=1
 
+RUN sed -i 's/granularity: microarchitectures/granularity: generic/' $SPACK_ROOT/etc/spack/defaults/concretizer.yaml
```
CodeGat, line 18: Hey, would you be able to give me an explanation of this line? I'm not too familiar with what its doing to the concretizer.

CodeGat (approved):
Just one comment for review, for my own understanding, that won't block approval. LGTM.

containers/Dockerfile.base-spack
harshula: https://spack.readthedocs.io/en/latest/build_settings.html#selection-of-the-target-microarchitectures

We had to change the default from microarchitectures because the instructions being used were architecture specific and different from the COSIMA build of ACCESS-OM2.

PR #66, opened 2023-08-22
Removed '@'s in package names

### LINKED ISSUE #65: Can't push package to `ghcr.io` as image name is invalid (has '@')
The current base-spack naming convention has an '@' between the compiler name and compiler version. '@' isn't actually allowed in an image name, so it must be renamed.

CodeGat:
'@'s aren't allowed in package names and would therefore fail to push. This PR fixes that. 

Should close #65 !


Quartzer2 (approved):
Looks good.

PR #81, opened 2023-08-24
Replace output comparison with exit code comparison

### LINKED ISSUE #80: Failing to kick of a build of `base-spack` when the image doesn't exist already
The `docker manifest inspect ...` command doesn't treat output like the other docker commands, namely, it seems to asynchronously output once it has contacted `ghcr.io` about the existence of an image.

CodeGat:
The `docker manifest inspect` command doesn't write to standard out in the expected way. So, instead of comparing the output of the command, we shall compare the exit code of the command (where nonzero means the image doesn't exist, and zero means it does).  

Should close #80 !


echus (approved):
Looks good :+1:

PR #82, opened 2023-08-24
dependency-image-pipeline now has access to the secrets of the caller workflow

CodeGat:
Callee workflows must inherit secrets from the caller workflow with `jobs.<job>.secrets.inherit`


CodeGat:
No longer relevant given the changes brought about by #84

PR #87, opened 2023-08-27
base-spack: specifiy target (x86_64)

harshula:


containers/Dockerfile.base-spack
```diff
@@ -13,6 +13,7 @@ ENV SPACK_PACKAGES_REPO_ROOT=/opt/spack_packages
 ENV SPACK_ENV_COMPILER_PACKAGE=${COMPILER_PACKAGE}
 ENV SPACK_ENV_COMPILER_VERSION=${COMPILER_VERSION}
 ENV SPACK_ENV_COMPILER_NAME=${COMPILER_NAME}
+ENV SPACK_ENV_TARGET=x86_64
```
CodeGat, line 16: It might be worth making this even more specific (eg. `SPACK_ENV_COMPILER_TARGET`)

CodeGat (changes requested):
Added one comment

containers/Dockerfile.base-spack
harshula: It's not just a compiler target. That's the target for everything. Later, it will likely end up being an `ARG`.

CodeGat: Gotcha, resolved.

CodeGat (approved):

Issue #86, opened 2023-08-25
Failing to push to buildcache as `intel-oneapi-compilers` matches multiple packages in the cache


CodeGat:
As noted in this run: https://github.com/ACCESS-NRI/build-ci/actions/runs/5972274202/job/16202688428#step:6:5971
```
"intel-oneapi-compilers" matches multiple packages: [
    intel-oneapi-compilers@=2021.1.2%gcc@=8.5.0+envmods build_system=generic arch=linux-rocky8-x86_64_v3, 
    intel-oneapi-compilers@=2021.1.2%gcc@=8.5.0+envmods build_system=generic arch=linux-rocky8-x86_64_v4]
```
The build then fails and the package is not pushed.

harshula:
We can use `target=x86_64` as an argument to all our `spack install` commands to ensure the target is constraint to only `x86_64`.

CodeGat:
Hey @harshula, but in this case, aren't they both `x86_64`, only differing by the version (`v3` vs `v4`?) Why would it be an error if there are two matching things being pushed to the buildcache (at lease, in terms of version)?

harshula:
Run `spack arch --known-targets`, that should make it clearer. i.e. x86_64 is not a superset.

harshula:
https://spack.readthedocs.io/en/latest/basic_usage.html#architecture-specifiers

harshula:
Try install the compiler with `target=x86_64`

harshula:
See if PR https://github.com/ACCESS-NRI/build-ci/pull/87 helps.

CodeGat:
Thanks, will do!

harshula:
Now delete all the compilers in the build cache and re-run. Fingers crossed.

CodeGat:
@harshula Similar failures occured, but I hadn't wiped the buildcache completely:
https://github.com/ACCESS-NRI/build-ci/actions/runs/5994625265/job/16256781237#step:6:2841
https://github.com/ACCESS-NRI/build-ci/actions/runs/5994625265/job/16256781096#step:6:5951

harshula:
[...]

PR #88, opened 2023-08-28
Added `target` to all other `spack install`s

CodeGat:
So spack uses the existing compiler target when in an environment rather than choosing for itself.

Should go some way to completing #86

### DIFF
--- a/containers/Dockerfile.base-spack
+++ b/containers/Dockerfile.base-spack
@@ -13,7 +13,7 @@ ENV SPACK_PACKAGES_REPO_ROOT=/opt/spack_packages
 ENV SPACK_ENV_COMPILER_PACKAGE=${COMPILER_PACKAGE}
 ENV SPACK_ENV_COMPILER_VERSION=${COMPILER_VERSION}
 ENV SPACK_ENV_COMPILER_NAME=${COMPILER_NAME}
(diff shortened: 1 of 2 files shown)


harshula (approved):

containers/Dockerfile.base-spack
```diff
@@ -67,7 +67,7 @@ RUN --mount=type=secret,id=access-nri.pub \
     spack gpg trust /run/secrets/access-nri.pub
 
 # Set up compilers for this base-spack image 
-RUN spack install ${SPACK_ENV_COMPILER_PACKAGE}@${SPACK_ENV_COMPILER_VERSION} target=${SPACK_ENV_TARGET}
+RUN spack install ${SPACK_ENV_COMPILER_PACKAGE}@${SPACK_ENV_COMPILER_VERSION} target=${SPACK_ENV_ARCH}
```
harshula, line 70: If you want to switch to arch, then you'll need to do arch=

harshula (changes requested):

containers/Dockerfile.base-spack
CodeGat: Noted, updated.

harshula (approved):

Issue #91, opened 2023-08-28
Update model CI to generically use `spack_packages` version


CodeGat:
Due to the updates to the CI pipeline, the model CI infrastructure  needs to have knowledge of the current `spack_packages` and compiler version of the dependency image.

CodeGat:
Issue closed as all the relevant PRs have been merged

PR #94, opened 2023-08-30
Make build-package workflow generic

CodeGat:
Most of the information in the model CI can be inferred from the model repository itself, or the `build-ci` repo.  For example: https://github.com/ACCESS-NRI/MOM5/blob/master/.github/workflows/test-build.yml contains information on the package name and compilers, which can be determined from the repository name and `build-ci/containers/compilers.json` respectively. 

This PR removes the inputs from the build-package reusable workflow and infers them in a `setup` job instead, which will allow all models to use the exact same workflow file. If they can all use the same workflow file, then we can create a starter workflow for any future models as well (provided there is a `.github` repo at the root of the organisation that specifies one).

References #91 (and will close it once all models implement the new generic build-packages)

### DIFF
--- a/.github/workflows/build-package.yml
+++ b/.github/workflows/build-package.yml
@@ -2,37 +2,68 @@ name: Build package
 
 on:
   workflow_call:
-    inputs:
-      container-registry:
-        required: true
-        type: string
-      container-name:
-        required: true
-        type: string
-      package-name:
-        required: true
-        type: string
-      compiler-name:
-        required: true
-        type: string
-      compiler-version:
-        required: true
-        type: string
-      spack-packages-version:
-        required: false
-        type: string
-        description: Either a git tag or branch name for the ACCESS-NRI/spack_packages repository, which defaults to the main branch
-        default: main
 
 jobs:
+  setup-spack-packages:
+    name: Get information from spack_packages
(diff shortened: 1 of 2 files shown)


.github/workflows/build-package.yml
```diff
@@ -2,37 +2,68 @@ name: Build package
...
+    - uses: actions/checkout@v3
+      with:
+        repository: access-nri/build-ci
+    - name: Get compilers to test
+      id: get-compilers
+      run: echo "compilers=$(jq -c . containers/compilers.json)" >> $GITHUB_OUTPUT
+    - name: Get model to test
+      id: get-model
+      # model-components are associated with an overarching model (for example, cice5 is associated with access-om2), this uses models.json to find the associated model
+      run: echo "model=$(jq -c 'to_entries[] | select(.value | contains([${{ github.event.repository.name }}]))' containers/models.json)" >> $GITHUB_OUTPUT
```
Quartzer2, line 38: As we discussed. There *might* be a better way of doing this. But I wouldn't worry. If it works, it works. Not like this will get a million hits an hour :)

Quartzer2 (approved):
Awesome. Thanks for running me through what you have done. 

All looks good!

Quartzer2 (approved):

PR #95, opened 2023-08-31
jq uses raw output and uses the lower-case repo name

CodeGat:
Failed to build image name because the output of jq was is in quotes, and image used the upper case MOM5. 
This PR make jq use the `-cr` flag (compressed and raw output) and also makes the setup-build-ci need setup-model for the lowercase model

References #91


jo-basevi (approved):
LGTM

PR #98, opened 2023-09-14
Standardized env and arg vars to use curly braces

### LINKED ISSUE #89: Standardise all environment variables on `${...}`
In all the Dockerfiles, standardise all `ENV`/`ARG` uses on `${...}` rather than `$...`. 
See here: https://stackoverflow.com/questions/8748831/when-do-we-need-curly-braces-around-shell-variables

CodeGat:
Instead of environment variables looking like `$THIS` they look like `${THIS}` now. Also changed sub-command expansion from backticks to `$(...)` as that is standard.

Should close #89

### DIFF
--- a/containers/Dockerfile.access-om2
+++ b/containers/Dockerfile.access-om2
@@ -3,13 +3,13 @@ FROM ${BASE_IMAGE}
 
 ARG PACKAGE_NAMES="libaccessom2 mom5 cice5 oasis3-mct"
 
-WORKDIR $SPACK_ROOT
+WORKDIR ${SPACK_ROOT}
 
 LABEL au.org.access-nri.ci.packages ${PACKAGE_NAMES}
 LABEL au.org.access-nri.ci.base-spack-image ${BASE_IMAGE}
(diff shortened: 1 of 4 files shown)


containers/Dockerfile.access-om2
```diff
@@ -7,9 +7,9 @@ WORKDIR $SPACK_ROOT
```
jo-basevi, line 7: Should the line above be `WORKDIR ${SPACK_ROOT}`? Sorry I couldn't select the correct line (line 6) in the review.

jo-basevi (approved):
There's also some backticks used around here - https://github.com/ACCESS-NRI/build-ci/blob/751a7f88ea60f0e1e2183070d745001bc950b300/containers/Dockerfile.base-spack#L122

Are you standardising the `$` to `${` through the whole repo or just the dockerfiles? Otherwise, it all looks good to me 👍

containers/Dockerfile.access-om2
CodeGat: Correct, good catch!

CodeGat (commented):
Added your suggestions. Although, if you add comments that you want reviewed, you don't have to approve the request :)

PR #99, opened 2023-09-18
JSON Schema and JSON Validation CI

### LINKED ISSUE #63: Lint/validate compilers.json
I should have picked this up when reviewing https://github.com/ACCESS-NRI/build-ci/pull/42, sorry, but we should be doing at least a linting check on https://github.com/ACCESS-NRI/build-ci/blob/main/containers/compilers.json

If we're storing important information in containers.json we should add a CI linting check when it is updated. It is quite easy to make a syntax error, especially with quoted strings.

There are some GitHub marketplace actions, but nothing maintained/well-used/appropriate so probably sufficient to just use jq in this case, so see https://github.com/jqlang/jq/issues/1637#issuecomment-653101641

or https://github.com/jqlang/jq/issues/1539 and check "array" is returned:

```bash
$ jq -c type compilers.json 
"array"
```

If we *really* wanted to check `compilers.json` file validity we would [create a JSON schema and validate against it](https://towardsdatascience.com/how-to-validate-your-json-using-json-schema-f55f4b162dce).

There is a marketplace action to do that, but it [doesn't seem to be actively maintained](https://github.com/OrRosenblatt/validate-json-action/issues/40)

https://github.com/marketplace/actions/validate-json

[There are lots of implementations](https://json-schema.org/implementations.html), for CLI though it is either [npm based](https://www.npmjs.com/package/ajv-cli) or [go based](https://github.com/neilpa/yajsv) which might just introduce a whole lot of dependency hell.

A non-CLI python option is https://horejsek.github.io/python-fastjsonschema/

CodeGat:
Added `*.schema.json` files for validation CI. Can be done on CLI with (for example) `jsonchema -i containers/compilers.json containers/compilers.schema.json`. 
Added build-ci CI (how meta!) for json validation. 

Should close #63 !

### DIFF
--- a/containers/compilers.schema.json
+++ b/containers/compilers.schema.json
@@ -1,0 +1,22 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "title": "Compilers",
+    "description": "A list of compilers that are supported",
+    "type": "array",
+    "items": {
+        "type": "object",
+        "properties": {
+            "name": {
+                "type": "string"
+            },
+            "package": {
+                "type": "string"
+            },
+            "version": {
+                "type": "string"
+            }
+        },
+        "required": [ "name", "package", "version" ],
+        "additionalProperties": false
+    }
+}
--- a/.github/workflows/json-lint.yml
+++ b/.github/workflows/json-lint.yml
@@ -1,0 +1,36 @@
+name: Validate JSON files
+on:
+  pull_request:
+    paths:
+      - '**.json' 
+jobs:
+  setup-validate-json:
+    name: Setup Validate JSON
+    runs-on: ubuntu-latest
+    outputs:
+      json-to-validate: ${{ steps.get-json.outputs.json }}
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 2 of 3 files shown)


CodeGat:
Woah, the CI json validations checks run automatically!
image

containers/compilers.schema.json
```diff
@@ -0,0 +1,21 @@
...
+                "type": "string"
+            },
+            "package": {
+                "type": "string"
+            },
+            "version": {
+                "type": "string"
+            }
+        },
+        "required": [ "name", "package", "version" ]
```
jo-basevi, line 19: Is it worth specifying `"additionalProperties": false` as well here? So if new fields are added in, the schema validation check will fail so at least it'll act as a reminder to update the schema..

jo-basevi (approved):
It looks good to me! Only have a small question, but it's probably fine as is 👍

containers/compilers.schema.json
CodeGat: Good idea!!

PR #107, opened 2023-09-27
Make JSON validation workflow reusable

### LINKED ISSUE #104: Generify `json-1-validate` so it can be used as a resuable workflow in other repos
The `json-1-validate` workflow is so close to being generic and reusable. Make it a callable reusable workflow for other repos that might make use of json validation.

Should be a relatively small fix(tm). 

References #100 (specifically https://github.com/ACCESS-NRI/build-ci/pull/100/files#r1336779968 )!

CodeGat:
Makes the JSON Validate workflow usable by:
* creating a new reusable workflow that has a `workflow_call` trigger (called `validate-json.yml`)
  * Remove references to directories (specifically `containers`) and referring to `workflow_call.inputs.src`
  * Expand all the flags of commands from their short version to the long version for more understandability (for example, `-R` to `--raw-input` for the call to `jq`)
* Simplifying the actual call in `json-1-validate.yml` to simply call the newly created workflow above. 

Closes #104 !

### DIFF
--- a/.github/workflows/validate-json.yml
+++ b/.github/workflows/validate-json.yml
@@ -1,0 +1,42 @@
+name: Validate JSON
+on:
+  workflow_call:
+    inputs:
+      src:
+        type: string
+        required: false
+        default: "."
+        description: "A directory that contains both the *.json/*.schema.json files (defaults to '.')"
+
+     
+jobs:
+  setup-validate-json:
+    name: Setup Validate JSON
+    runs-on: ubuntu-latest
+    outputs:
+      json-to-validate: ${{ steps.get-json.outputs.json }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Get all schema 
+        id: get-json
+        # this command finds all the filenames with the '.schema.json' extension, removes that extension, then turns them into a json array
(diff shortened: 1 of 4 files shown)


.github/workflows/validate-json.yml
```diff
@@ -0,0 +1,42 @@
...
+      fail-fast: false
+      matrix: 
+        file: ${{ fromJson(needs.setup-validate-json.outputs.json-to-validate) }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
+        with:
+          python-version: '3.10'
+      - name: Validate files
+        run: jsonschema --instance ${{ inputs.src }}/${{ matrix.file }}.json ${{ inputs.src }}/${{ matrix.file }}.schema.json
```
aidanheerdegen, line 41: Comment to future self (and anyone else listening in), `jsonschema` is from the [python-jsonschema package](https://github.com/python-jsonschema/jsonschema) which is installed with `pip` in the `actions/setup-python@v3` workflow.

The code for the CLI that is being used is here:

https://github.com/python-jsonschema/jsonschema/blob/main/jsonschema/cli.py

aidanheerdegen (approved):
LGTM

PR #122, opened 2023-11-21
Dockerfile.base-spack: explicitly install cmake 3.24.2

harshula:
* The latest cmake version in Gadi's /apps is 3.24.2


CodeGat (approved):
LGTM

PR #123, opened 2023-11-22
Dockerfile.base-spack: clone a spack release branch instead of a tag

harshula:
Cloning a release branch enables important bug fixes, via minor releases, to be automatically received without having to update this Dockerfile after each minor release creates a new tag.


CodeGat (approved):
Ah, good idea. Approved

aidanheerdegen (approved):
Nice catch!

PR #124, opened 2023-11-23
Use ACCESS-NRI/docker-build-push action

CodeGat:
In this PR:
* We delete the `build-docker-image.yml` workflow and use our own `access-nri/actions/.github/actions/docker-build-push` action instead.
* A _bunch_ of `README.md` changes.

As a cascading effect, since we don't have to run the now-deleted reusable workflow as a job, and instead can run the docker building and pushing as a step, we can greatly simplify the overall workflow. Specifically:
* `dep-3-image-build.yml`: We can delete this workflow as it simply sets up and builds the dependency image in a workflow - these can become steps in the calling workflow instead. 
* `dep-2-build-base.yml`: This can be renamed `dep-2-build.yml` as it builds both the `base-spack` and `dependency` images. 
* `dep-2-build.yml`: 
  * As noted earlier, the steps in `dep-3-image-build` can now become steps in the dependency-image job, and we can further merge steps together
  * Now that the `base-spack` job also uses the custom action, it can be simplified into a series of steps rather than a series of jobs.


harshula (approved):
Reviewed in VC.

PR #138, opened 2024-01-11
Dependency Image Spack fix

### LINKED ISSUE #137: Testing of Models failing due to built Dependency Image spack errors
The model workflows that call `model-1-build.yml` are failing because GitHub actions doesn't support Dockerfile-defined `ENTRYPOINT`s. This leads to spack not initializing correctly, and therefore the entire build failing.

CodeGat:
In this PR:
* `Dockerfile.dependency`: 
  * Removed garbage collection as it was deleting packages that would be reinstalled later
  * Removed `ENTRYPOINT` command and added a comment noting why it isn't there (see: https://stackoverflow.com/questions/66358482/github-actions-ignores-overrides-docker-containers-entrypoint)
* `model-1-build.yml`: 
  * Added setup of environment and compilers when running the container
  * Added volumes that store various environment files that are uploaded as artifacts
  * Removed `tmate` remote debugging because the logs should be sufficient 
  * Cleaned up some of the code and logic

Closes #137

### DIFF
--- a/.github/workflows/model-1-build.yml
+++ b/.github/workflows/model-1-build.yml
@@ -117,6 +117,10 @@ jobs:
     - name: Generate force-concretized lockfile
       if: failure()
       run: |
+        . $SPACK_ROOT/share/spack/setup-env.sh
+        spack load ${{ matrix.compiler.package }}@${{ matrix.compiler.version }}
+        spack compiler find
+        spack env activate ${{ env.PACKAGE_NAME }}
         spack concretize --force
         cp ${{ env.SPACK_YAML_LOCATION }}/spack.lock /lockfiles/force.spack.lock
 
--- a/.github/workflows/model-1-build.yml
+++ b/.github/workflows/model-1-build.yml
@@ -110,7 +110,7 @@ jobs:
         cp ${{ env.SPACK_YAML_LOCATION }}/spack.yaml /lockfiles/generic.spack.yaml
         cp ${{ env.SPACK_YAML_LOCATION }}/spack.lock /lockfiles/generic.spack.lock
         echo "------------------------------------------------------------------------------"
-        spack change --match-spec ${{ env.PACKAGE_NAME }} ${{ env.PACKAGE_NAME }}@git.$GH_REF%intel@2021.2.0 arch=linux-rocky8-x86_64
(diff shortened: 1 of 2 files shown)


.github/workflows/model-1-build.yml
```diff
@@ -90,49 +100,38 @@ jobs:
...
-
-    - name: Generate lockfile
+    - name: Build ${{ env.PACKAGE_NAME }} via spack
       run: |
-        cd $SPACK_ROOT/var/spack/environments/${{ needs.setup-model.outputs.package-name }}
-        cp spack.lock ${{ needs.setup-model.outputs.package-name }}.original.spack.lock
+        . $SPACK_ROOT/share/spack/setup-env.sh
+        spack load ${{ matrix.compiler.package }}@${{ matrix.compiler.version }}
+        spack compiler find
+        spack env activate ${{ env.PACKAGE_NAME }}
```
aidanheerdegen, line 108: In the quest to have an easy way to test CI locally with the docker images, it would be good if the commands required to reproduce the CI environment were easily reproducible.

To that end, would it make sense to put them in a separate script in the repo? That would require passing it some variables, which might be a pain and so undesirable.

Another option would be to echo the commands necessary to reproduce the environment when the CI fails. That way someone can look at the logs and copy and paste the exact commands required. Not sure if there is a nice way to do that once and reuse it though.

aidanheerdegen (approved):
Looks good, just wondering about providing a nice way to reproduce exact environment when debugging CI issues.

.github/workflows/model-1-build.yml
CodeGat: In the logs, I believe it substitutes the values noted (like `matrix.compiler.package` and `env.PACKAGE_NAME`).

.github/workflows/model-1-build.yml
```diff
@@ -90,49 +100,38 @@ jobs:
...
+        spack find --show-concretized --long
+        cp ${{ env.SPACK_YAML_LOCATION }}/spack.yaml /lockfiles/generic.spack.yaml
+        cp ${{ env.SPACK_YAML_LOCATION }}/spack.lock /lockfiles/generic.spack.lock
+        echo "------------------------------------------------------------------------------"
+        spack change --match-spec ${{ env.PACKAGE_NAME }} ${{ env.PACKAGE_NAME }}@git.$GH_REF%intel@2021.2.0 arch=linux-rocky8-x86_64
```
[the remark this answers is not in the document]
CodeGat: Nope! Updated.

aidanheerdegen (approved):
LGTM

PR #140, opened 2024-01-17
Dockerfile.base-spack: remove cmake version from dev target

harshula:
The CMake version will now come from the spack-config repository. See https://github.com/ACCESS-NRI/spack-config/pull/21 .


CodeGat (approved):
Looks good to me! Good to get this stuff out of the dockerfile and into spack-config

PR #141, opened 2024-01-18
Dockerfile.base-spack: add new args MPI_NAME and MPI_VERSION

harshula:
The MPI name and version should not be hardcoded in the dev target.

### DIFF
--- a/containers/Dockerfile.base-spack
+++ b/containers/Dockerfile.base-spack
@@ -138,12 +138,15 @@ RUN --mount=type=secret,id=access-nri.pub \
 ################################################################################
 FROM base-spack as dev
 
+ARG MPI_NAME=openmpi
+ARG MPI_VERSION=4.0.2
+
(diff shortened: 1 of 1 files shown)


containers/Dockerfile.base-spack
```diff
@@ -60,6 +60,8 @@ ARG SPACK_ARCH=linux-rocky8-x86_64
 ARG COMPILER_PACKAGE=intel-oneapi-compilers
 ARG COMPILER_VERSION=2021.2.0
 ARG COMPILER_NAME=intel
+ARG MPI_NAME=openmpi
+ARG MPI_VERSION=4.0.2
```
CodeGat, line 64: If these `ARGS` are only used in the `dev` image, perhaps they should be defined there instead - after the `FROM base-spack as dev` part, maybe?

CodeGat (commented):
Just one comment!

CodeGat (approved):

PR #143, opened 2024-01-18
Dockerfile.base-spack: explicitly install perl in base-os

harshula:
Perl is installed implicitly as a dependency in target base-os. Make it explicit now that we do not build it via Spack.
See https://github.com/ACCESS-NRI/spack-config/pull/22

### DIFF
--- a/containers/Dockerfile.base-spack
+++ b/containers/Dockerfile.base-spack
@@ -35,6 +35,7 @@ RUN dnf update -y \
     make \
     patch \
     patchutils \
+    perl \
     pkgconf \
     pkgconf-m4 \
     pkgconf-pkg-config \


aidanheerdegen (approved):
My only comment is we're specifying the exact perl version in the `spack` config, but not here. Could we end up with a mis-match, so `spack` doesn't find the external and decides to build it?

harshula:
Thanks! Perhaps we should remove the version number in both ci & gadi's `packages.yaml` in `spack-config`. In both instances, it is the system perl version and we let the operating system decide on the version.
However, the inclusion of the version number should not impact Spack's ability to find it. `spec` is something we define to describe the system perl, as per https://spack.readthedocs.io/en/latest/packages_yaml.html .

PR #144, opened 2024-01-21
Dockerfile.base-spack: check if $SPACK_CONFIG_DIR exists

harshula:
While attemping to test a new version of Spack via the Docker image, I noticed that $SPACK_ROOT/etc/spack didn't contain symlinks to spack-config config files. This change will generate an ENOENT error, if the directory for a particular Spack version does not exist.


CodeGat (approved):
Looks good to me :)

PR #150, opened 2024-03-18
compilers.json: Removed intel@2021.4.0

CodeGat:
See https://github.com/ACCESS-NRI/build-ci/actions/runs/8291385383/job/22690995420#step:4:3320

Due to this compiler failing to be fetched from intels sources, we have removed it and replaced it with `intel@2021.4.0`.

### DIFF
--- a/config/compilers.json
+++ b/config/compilers.json
@@ -2,11 +2,6 @@
   {
     "name": "intel",
     "package": "intel-oneapi-compilers",
-    "version": "2021.2.0"
-  },
-  {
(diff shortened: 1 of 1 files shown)


aidanheerdegen (approved):
The 2021.2.0 compiler seems pretty broken/not available, and as we have no idea when/if it will be fixed it is best to move to a compiler version we know is available. As we're just doing build CI checks I don't think it matters a great deal that we're changing versions, and if it did we should be worried about that!

LGTM in other words

PR #157, opened 2024-05-08
`models.json` Entry for ACCESS-ESM1.5

CodeGat:
Addition of components for `access-esm1.5`. For now, this will only be the public components as the docker images are freely pullable. Might have to turn the packages into `internal` packages. 

In this PR:
* models.json: Added public access-esm1.5 package components


aidanheerdegen (approved):
LGTM.

I hadn't thought about the docker images. Good catch.

CodeGat:
References ACCESS-NRI/cice4#4

PR #164, opened 2024-06-13
Dockerfile.base-spack: remove unnecessary ENV variables

harshula:
* Add label spack-config-repo-version

### DIFF
--- a/containers/Dockerfile.base-spack
+++ b/containers/Dockerfile.base-spack
@@ -59,6 +59,7 @@ ARG SPACK_REPO_VERSION=releases/${SPACK_VERSION}
 ARG SPACK_PACKAGES_REPO_VERSION=main
 ARG SPACK_CONFIG_REPO_VERSION=main
 ARG SPACK_TARGET=x86_64
+ARG SPACK_CONFIG_DIR=/opt/spack-config/${SPACK_VERSION}/ci
 ARG COMPILER_PKG_NAME=intel-oneapi-compilers
 ARG COMPILER_PKG_VERSION=2021.2.0
 ARG COMPILER_NAME=intel
@@ -67,9 +68,6 @@ ARG COMPILER_VERSION=2021.2.0
 ENV SPACK_ROOT=/opt/spack
 ENV GNUPGHOME=${SPACK_ROOT}/opt/spack/gpg
 ENV ENV_SPACK_TARGET=${SPACK_TARGET}
-ENV SPACK_PACKAGES_REPO_ROOT=/opt/spack-packages
-ENV SPACK_CONFIG_REPO_ROOT=/opt/spack-config
-ENV SPACK_CONFIG_DIR=${SPACK_CONFIG_REPO_ROOT}/${SPACK_VERSION}/ci
 ENV ENV_COMPILER_PKG_NAME=${COMPILER_PKG_NAME}
 ENV ENV_COMPILER_PKG_VERSION=${COMPILER_PKG_VERSION}
 ENV ENV_COMPILER_NAME=${COMPILER_NAME}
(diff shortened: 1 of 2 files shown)


containers/Dockerfile.base-spack
```diff
@@ -59,37 +59,31 @@ ARG SPACK_REPO_VERSION=releases/${SPACK_VERSION}
...
+LABEL au.org.access-nri.image.arch ${SPACK_ARCH}
 
 SHELL ["/bin/bash", "-c"]
 
 # Install spack
 RUN git clone -c feature.manyFiles=true ${SPACK_GIT_URL} ${SPACK_ROOT} --branch ${SPACK_REPO_VERSION} --single-branch --depth=1
 
 # Install ACCESS-NRI's spack-packages repo
-RUN git clone  https://github.com/ACCESS-NRI/spack-packages.git ${SPACK_PACKAGES_REPO_ROOT} --branch ${SPACK_PACKAGES_REPO_VERSION}
+RUN git clone  https://github.com/ACCESS-NRI/spack-packages.git /opt/spack-packages --branch ${SPACK_PACKAGES_REPO_VERSION}
```
CodeGat, line 83: Why do we need to remove the `SPACK_PACKAGES_REPO_ROOT` (and `SPACK_CONFIG_REPO_ROOT`)? Couldn't it just be converted into an `ARG` rather than an `ENV`, if we don't want the variable after the build?
It might be unlikely to have a `spack-{packages,config}` somewhere other than `$SPACK_ROOT/..`, but it may happen one day, and we might like to have that customisation not hard-coded in the Dockerfile. What do you reckon?

containers/Dockerfile.base-spack
```diff
@@ -59,37 +59,31 @@ ARG SPACK_REPO_VERSION=releases/${SPACK_VERSION}
...
 ARG SPACK_ARCH=linux-rocky8-x86_64
+ARG SPACK_CONFIG_DIR=/opt/spack-config/${SPACK_VERSION}/ci
 ARG COMPILER_PKG_NAME=intel-oneapi-compilers
 ARG COMPILER_PKG_VERSION=2021.2.0
 ARG COMPILER_NAME=intel
 ARG COMPILER_VERSION=2021.2.0
 
 ENV SPACK_ROOT=/opt/spack
 ENV GNUPGHOME=${SPACK_ROOT}/opt/spack/gpg
-ENV ENV_SPACK_ARCH=${SPACK_ARCH}
```
CodeGat, line 69: There is a single usage of this `ENV` in the `model-1-build.yml` workflow, here: https://github.com/ACCESS-NRI/build-ci/blob/3c0840d775d8f3d67cfeedb44173caa2682fa27e/.github/workflows/model-1-build.yml#L113. We'd need to get this information from somewhere else, in that case, or remove the `arch` specifier.

CodeGat (commented):
Just one comment. Generally, I agree with removal of the `ENV`s if they aren't needed.

containers/Dockerfile.base-spack
harshula: `/opt/spack-packages` can't be changed, it's also hardcoded in `spack-config` repo's `repos.yaml`.

containers/Dockerfile.base-spack
harshula: Good catch! Do you have preference?

harshula: Good catch! Do you have a preference?

harshula: The plan is to:
1. Try using `target` instead of `arch`.
2. Then read the `target` from `compilers.json`.

harshula: Hi @CodeGat , Let me know when the PR to read `target` from `compilers.json`, is ready for review.

CodeGat: I take it back, there is a few references to the compiler environment variables here:
https://github.com/ACCESS-NRI/build-ci/blob/3c0840d775d8f3d67cfeedb44173caa2682fa27e/containers/setup-spack-envs.sh#L10-L13
I'll see if I can get that information from the workflow I suppose.

CodeGat:
@harshula shouldn't this be `0.21` rather than `0.22`? Unless you want `build-ci` to by default be the latest `spack` release?

harshula:
Hi @CodeGat , No, build-ci should be using the Spack version set in its new config file and NOT using the default value in the Dockerfile.

containers/Dockerfile.base-spack
harshula: Changed the commits to keep the ENV variables that are needed.

CodeGat (approved):
Looks good. Thanks for keeping some of the ones that are used later in (like `ENV_SPACK_TARGET`.

PR #172, opened 2024-08-12
Dockerfile.base-spack: update os and spack versions

harshula:


harshula:
Instructions for pulling RL has changed here: https://hub.docker.com/r/rockylinux/rockylinux

CodeGat (approved):
Excellent idea! We should come up with a checklist of TODOs for when we update to a new `spack` major...

PR #173, opened 2024-08-12
Dockerfile.base-spack: use target instead of arch

harshula:


CodeGat (approved):
Verified that there are no other references to the old `SPACK_ARCH` variable. Thanks for the update.

harshula:
Test build completed successfully!

PR #175, opened 2024-08-21
Dockerfile.base-spack: keep in-sync with spack-config/spack-enable.bash

harshula:


containers/Dockerfile.base-spack
```diff
@@ -73,6 +73,10 @@ ENV ENV_COMPILER_PKG_VERSION=${COMPILER_PKG_VERSION}
 ENV ENV_COMPILER_NAME=${COMPILER_NAME}
 ENV ENV_COMPILER_VERSION=${COMPILER_VERSION}
 
+# Keep in-sync with spack-config/spack-enable.bash
+ENV SPACK_DISABLE_LOCAL_CONFIG="true"
+ENV SPACK_USER_CACHE_PATH="/opt"
```
CodeGat, line 78: Can we do `$SPACK_ROOT/../` or is that ENV not available to us?

harshula: Good idea!

CodeGat (approved):

PR #187, opened 2024-12-16
Support non-default RHS versions, variants

CodeGat:
In this PR:
* Add new inputs to the model build stage that allow customizable variants/RHS version sections. This is left to the caller to implement.


jo-basevi (approved):
LGTM

PR #189, opened 2025-02-06
Dockerfile.base-spack: build gmake before cmake

harshula:
* This ensures that cmake depends on the gmake built with the compiler we selected.


CodeGat (approved):
Fair enough!

PR #196, opened 2025-03-11
CI 2.0: Reworked Dockerfile, Added Compose and Config

### LINKED ISSUE #194: Update Docker Image to use Entrypoint Inputs
References https://github.com/ACCESS-NRI/build-ci/issues/191

## Background

Since we are moving towards a more customizable image, we need to be able to ingest `spack.yaml`s and build from them. Depending on the above issue, we also may need to build with different compilers/etc. This requires changes to the existing dockerfiles, and may require things like volumes/docker compose.

CodeGat:
Closes #194

## Background

This PR updates the docker infrastructure for an upcoming Build CI 2.0 update. There may have to be further alterations as more requirements come in (for example, where tests are stored, and the exact location of buildcaches). 

The goal of this PR is to get development infrastructure as close to how it will be in production:

| Difference | Production | Development (This PR) |
| ---------- | ---------- | --------------------- |
| Orchestrated Via | Kubernetes Cluster (Using `kompose`) | Docker Compose |
| Running On | Self-hosted, ephemeral GitHub Actions Runners via Actions Runner Controller | Locally Via Linked Docker Containers |
| Upstream Spack | A Pod within the Cluster that contains an instance of `spack` that has compilers and low-level packages installed | A Service container that contains the compilers and low-level packages installed |
| Downstream Runners | ARC Runners will connect to the Upstream with a ReadOnlyMany volume, which allows them to build quicker | The Downstream Spack will connect to the upstream via a read only volume |

The Dockerfile has been simplified, and now has two main targets after `base-spack` - `runner` (for downstream `spack`s to install and test packages) and `upstream` (a compiler-and-package store). 
Docker compose files have been added, for both development (`dev.compose.yaml`) and production (`compose.yaml`, in which `kompose` will generate the equivalent deployment objects). 

A bunch of files that aren't relevant to the 2.0 have been removed, also. 

## The PR

In this PR:

- **Update Dockerfile to use runner and upstream targets**
- **Delete Dockerfile.[base-spack|dependency] as they are consolidated into Dockerfile**
- **Add `containers/config` for packages to install upstream**
- **Delete [models|compilers][.schema].json, superseded by packages.json**
- **Delete setup-spack-envs.bash, not required in current solution**
- **Add production and development compose file**
- **Add ci-spack-enable.bash for loading and finding compilers downstream**

## Testing

Tested via running the following (among other things...): 
```bash
COMPOSE_BAKE=true docker compose -f containers/dev.compose.yaml up --detach --build
docker compose -f containers/dev.compose.yaml exec runner /bin/bash
# Inside the runner container...
. /opt/spack-config/ci-spack-enable.bash
# You now have the compilers loaded from upstream, and can install stuff using them!
spack install datetime-fortran%intel@2021.2.0 target=x86_64
```


containers/config/dev.packages.json
```diff
@@ -0,0 +1,11 @@
+{
+    "$schema": "./packages.schema.json",
+    "packages": [
+        "gmake%intel@2021.2.0 target=x86_64",
+        "cmake%intel@2021.2.0 target=x86_64",
+        "openmpi@4.0.2%intel@2021.2.0 target=x86_64"
```
aidanheerdegen, line 6: I think we probably want to use `openmpi@4.1.5` as I think that is where @harshula is heading.

containers/config/packages.json
```diff
@@ -0,0 +1,10 @@
+{
+    "$schema": "./packages.schema.json",
+    "packages": [
+
+    ],
+    "compilers": [
+        "intel-oneapi-compilers@2021.2.0 target=x86_64",
+        "intel-oneapi-compilers@2021.4.0 target=x86_64"
```
aidanheerdegen, line 8: I think we should have a classic and non-classic here. We're going to want to test both at the moment.

containers/config/dev.packages.json
```diff
@@ -0,0 +1,11 @@
...
+    ],
+    "compilers": [
+        "intel-oneapi-compilers@2021.2.0 target=x86_64"
```
aidanheerdegen, line 9: I don't think I understand the different use cases for production and dev, so hard to comment on which compilers should be included here.

containers/config/packages.json
```diff
@@ -0,0 +1,10 @@
+{
+    "$schema": "./packages.schema.json",
+    "packages": [
+
```
aidanheerdegen, line 4: Is this empty intentionally?

containers/dev.compose.yaml
```diff
@@ -0,0 +1,38 @@
...
+      context: .
+      target: upstream
+      args:
+        OS: rocky
+        SPACK_VERSION: v0.22
+        SPACK_CONFIG_REPO_VERSION: 2025.03.0
+        SPACK_CONFIG_DIR: /opt/spack-config/v0.22/ci-upstream
+        SOURCE_PACKAGES_JSON_FILE: config/dev.packages.json
+    volumes:
+      - dev-upstream:/opt/upstream:rw
```
aidanheerdegen, line 17: Is there a reason to call this `dev-upstream`? Seems unnecessarily different, unless the dev and release containers have to be mounted a the same time?

containers/spack-config/ci-spack-enable.bash
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+# !!! This is not an executable! Source this file !!!
```
aidanheerdegen, line 3: If we don't want to execute this file [this trick is cool](https://unix.stackexchange.com/a/424495)

```bash
if [ "${BASH_SOURCE[0]}" -ef "$0" ]
then
    echo "Hey, you should source this script, not execute it!"
    exit 1
fi
```

As well as [setting the shebang line to `#!/bin/false`](https://unix.stackexchange.com/a/424498)

aidanheerdegen (commented):
Note that I have not run the container commands, just had a gander at the source

containers/Dockerfile
```diff
@@ -53,11 +54,12 @@ RUN dnf update -y \
...
  && rm -rf /var/cache/dnf \
- && dnf clean all
-
+ && dnf clean all \
+ && curl -L https://github.com/mikefarah/yq/releases/download/v4.45.1/yq_linux_amd64 -o /usr/bin/yq \
+ && chmod +x /usr/bin/yq
 
 ################################################################################
-FROM opensuse/leap:15.5 as base-os-suse
+FROM opensuse/leap:15.5 AS base-os-suse
```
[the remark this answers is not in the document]
CodeGat: That is a @harshula question since he added it...I think it was because Pawsey uses Leap, and he could create a dev environment based on that OS.

containers/config/dev.packages.json
CodeGat: Essentially `dev.packages.json` is for local `docker compose` builds (analogous to what @harshula was doing on his own machine, having some packages installed earlier for his own testing). `packages.json` is for what will be on the kubernetes upstream Pod. Should they be different packages/compilers?

containers/config/dev.packages.json
CodeGat: Done in https://github.com/ACCESS-NRI/build-ci/pull/196/commits/ce9c12b42cdf5baa980b50433d20dc894a0410e7

containers/config/packages.json
CodeGat: For now, yeah. I'm not too sure about what packages we would want to install upstream in production...

containers/dev.compose.yaml
CodeGat: For local testing of `dev.compose.yaml` and `compose.yaml` I didn't want the volumes to be shared lest I get confused about what is mounted :sweat_smile:

containers/config/packages.json
CodeGat: Fair enough! It probably should be a wider discussion about what we want installed on the production upstream pod...

containers/spack-config/ci-spack-enable.bash
CodeGat: Good idea, done in https://github.com/ACCESS-NRI/build-ci/pull/196/commits/9b8bb86a7e3002cbd1796abcfd9eaa0d12189f6b

containers/Dockerfile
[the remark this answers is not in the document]
harshula: It's for Pawsey.

containers/config/packages.json
harshula: Add:
intel-oneapi-compilers@2023.2.4
intel-oneapi-compilers@2025.0.4

containers/Dockerfile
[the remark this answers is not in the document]
aidanheerdegen: Cool, just wanted to make sure we needed it.

containers/config/packages.json
aidanheerdegen: I think a wider conversation is required, but we should plough ahead with something representative for the time being.

containers/config/dev.packages.json
aidanheerdegen: > Should they be different packages/compilers?

Case for yes: it is the same as production, so tests are representative of that
Cade for no: takes too long and is too large for testing. Testing also likely to want latest rather than re-run old tried and tested.

I think I'd come down on the "no" side: have "representative" compilers in `dev` + bleeding-edge latest that might not be in release.

containers/config/packages.json
aidanheerdegen: Can we come up with an initial list to get on with? A few flavours of CMake (stable + latest?), OpenMPI (can still build their own, but have some vanilla options available for speed), ditto for netCDF and ESMF?

I'm thinking as we're going to have many more CI builds it would be good to have some widely used software packages available as spack packages to speed up the build process.

containers/dev.compose.yaml
aidanheerdegen: And that is a good reason.

containers/config/packages.json
CodeGat: Done in the most recent commits.

containers/config/dev.packages.json
CodeGat: Done  in later commits :)

containers/config/packages.json
CodeGat: Wider conversation happened, and the feedback implemented in later commits

aidanheerdegen (approved):
LGTM.

Note for later (maybe when compilers are just normal packages), but can we move from a `packages.json` to utilising a `spack.yaml` environment file to specify the packages we want installed in the base image? It would make it a lot easier to specify compiler, target and packages in a matrix.

CodeGat:
That's a good point. I would have to see if the packages/compilers are accessible outside of the environment. I actually might make this change now...

CodeGat:
I think that one of the maybe-issues with the `spack.yaml` solution is that we couldn't install the compilers and packages together in the same `spack.yaml`, because we need to `spack load <compiler> ` and `spack compiler find` before the packages are done. I'll just confirm that.

CodeGat:
In any case, the `spack.yaml` would look something like this:
```
spack:
  definitions:
  - compiler_packages:
    - intel-oneapi-compilers@2023.2.4 target=x86_64
    - intel-oneapi-compilers@2025.0.4 target=x86_64

  - compilers:
    - intel@2021.10.0 target=x86_64
    - intel@2025.0.4 target=x86_64

  - packages:
    - gmake
    - cmake
    - openmpi@4.1.5

  - specs:
    - matrix:
      - $packages
      - $%compilers

  specs:
  - $compiler_packages
  - $specs  
```
Which I agree looks much cleaner...

CodeGat:
Unfortunately @aidanheerdegen, the above doesn't work. When it is concretizing the spec, it can't find the compilers for the packages, because the package_compilers are not yet installed, so it fails. The only way to do it so far would be to have two (actually four, if we have `dev`/`prod`) manifests, in which one is the compiler packages installed first, and then the other is the actual packages.

aidanheerdegen:
> `spack v1.0.0` might be a different story. But for now I'll leave it as it is.

Cool. Good to have that in mind for the future.

PR #198, opened 2025-03-18
Dockerfile.base-spack: update to openmpi 4.1.5 and latest ifort

harshula:
* Also delete files in sourcecache to reduce the image size.


harshula:
Test build still running ...

containers/Dockerfile.base-spack
```diff
@@ -213,5 +213,7 @@ RUN spack install \
     ${MPI_NAME}@${MPI_VERSION} %${COMPILER_NAME}@${COMPILER_VERSION} \
     target=${SPACK_TARGET}
 
+RUN spack clean --downloads
```
CodeGat, line 216: As a further space saving measure - you should put the install commands and spack clean --downloads as part of the same RUN command. Having multiple installation layers (and then a clean up layer) probably won't save as much space as you can, as layer creation has space overhead. 

Essentially, when you build or pull this image down, you will download a layer where the downloads are all still there, and then download a layer that deletes them. 

 Something more like:
```
RUN <<EOF
spack install gmake %${COMPILER_NAME}@${COMPILER_VERSION} target=${SPACK_TARGET}
spack install cmake %${COMPILER_NAME}@${COMPILER_VERSION} target=${SPACK_TARGET}
spack install ${MPI_NAME}@${MPI_VERSION} %${COMPILER_NAME}@${COMPILER_VERSION} target=${SPACK_TARGET}
spack clean --downloads
EOF
```
would be more space efficent.

CodeGat: Badly drawn explanation incoming...
image

Each RUN step is a layer.
Essentially when pulling down that image, you will have to pull down layers that has a single install, and the downloads that we will remove later. Only at the end do we clean up the filesystem, but the overall image size is still relatively large due to those earlier layers. 

If we do all the install and cleanup in a single layer, the layers themselves are also optimized, since we don't store the downloads in an earlier layer.

harshula: That change had negligible impact. However, doing a `spack clean --downloads` immediately after `spack install ${COMPILER_PKG_NAME}@${COMPILER_PKG_VERSION} target=${SPACK_TARGET}`  saved more than 2 GiB with the ifort compiler.

harshula:
### Successful builds:
spack-dev-suse    v0.22.5-intel-20250318    61f4a964ad55   6 days ago     6.66GB
spack-dev-rocky   v0.22.5-oneapi-20250318   1f9595d0b5a5   6 days ago     4.63GB
spack-dev-rocky   v0.22.5-intel-20250318    59e8b6ee4dee   6 days ago     6.46GB

CodeGat (approved):
Looks like some good space-saving. Thanks!

PR #200, opened 2025-03-25
Build-ci 2.0 Entrypoint Workflow

### LINKED ISSUE #190: Consider Entrypoint Workflow Requirements
In a similar vein to `build-cd`, we need to decide how Model Component Repositories (MCRs?) will interface with an updated `build-ci`. 

Usually, we do this via Entrypoint Workflows, which are `on.workflow_call`-triggered workflows stored in `build-ci`. Also consider using Custom Actions. 

## Considerations

* An entrypoint workflow should contain a minimal amount of required arguments (preferably getting input via files stored on the Model Component Repository), defaults of which should be handled by the entrypoint workflow. This is so there are minimal major version updates to Model Component Repositories when the Entrypoint Workflows change. 
* An Entrypoint workflow should consider multiple types of invocation - `on.pull_request`, `on.push`, `on.workflow_dispatch`...even `on.repository_dispatch` or webhook event. Usually, we have enough information within the existing contexts to handle what must be done for different triggers. Consider a `defaults` job that sets the required inputs regardless of trigger event.
* Output important information from these jobs. Don't discount the possibility that MCRs will have custom logic post-entrypoint workflow.

### LINKED ISSUE #199: Investigate graceful failure of non-installed compilers in spec
We have compilers installed in the upstream image, but this is not all the possible compilers to install. Users may want to install different compilers. 

Currently in spack, if a compiler is not found or the compiler package is not installed, the installation fails. We need a way to install a compilers associated compiler package, and then attempt to re-install  the original spack manifest. 

## Implementation Details

* spack has deprecated the use of `config.install_missing_compilers`, which would have fixed this.
* We need a way to enumerate the compiler packages that would provide a given compiler - some are not obvious (e.g. `intel-oneapi-compilers@2023.2.4` provides `%intel@2021.10.0`)
* spack v1.0.0 treats compilers as packages - hopefully it would install them automatically and link them. 

First Pass Solution:

* Attempt installation of spack manifest
* If it fails the first time, check if that is due to a compiler not being installed
* Consult the enumeration of compiler packages and `spack install --add COMPILER_PACKAGE`
* Run `spack load COMPILER_PACKAGE && spack compiler find`
* Attempt to reinstall the original spack manifest 

Second Pass Solution:

* Let users add a compilers-only manifest to the workflow, which installs and links compilers before installing the actual manifest.

CodeGat:
Closes #190
Closes #199 

## Background

One of the most important things to get right for `build-ci 2.0` is to have appropriate workflow inputs with sensible defaults, and to only have required inputs when needed - this is to make it so there are only rare cases where all model component repositories CI have to be updated, because it's a bit of a pain!

The workflow steps are only guidance of how the inputs are going to be used, and there are a lot of `FIXME`s that will be resolved in later pull requests. This PR is to gather feedback on the inputs that all model component repositories will have to provide. 

## The PR

- **Add entrypoint inputs, stub of workflow**
- **Export environment variables from container, convert existing checkout logic to action**
- **Add future fixmes to code**


.github/workflows/ci.yml
```diff
@@ -0,0 +1,143 @@
...
+      - name: Export environment variables into GitHub Actions format
+        # Environment variables inside containers are not accessible in the `env` context,
+        # a context which would be used in future conditional steps.
+        # This step exports important container environment variables as outputs.
+        id: env
+        run: |
+          echo "SPACK_ROOT=$SPACK_ROOT" >> $GITHUB_OUTPUT
+          echo "SPACK_VERSION=$SPACK_VERSION" >> $GITHUB_OUTPUT
+          echo "SPACK_CONFIG_REPO_VERSION=$SPACK_CONFIG_REPO_VERSION" >> $GITHUB_OUTPUT
+          echo "SPACK_PACKAGES_REPO_VERSION=$SPACK_PACKAGES_REPO_VERSION" >> $GITHUB_OUTPUT
```
aidanheerdegen, line 80: Might be good to have this as an action to make the logic clearer for dummy users like me.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,143 @@
...
+        with:
+          repository-path: ${{ steps.env.outputs.SPACK_ROOT }}
+          ref: ${{ inputs.spack-ref }}
+
+      - name: Relink spack-config to spack
+        if: steps.spack-config-update.conclusion == 'success' || steps.spack-update.conclusion == 'success'
+        run: ln --symbolic --relative --verbose --force ${SPACK_CONFIG_DIR}/* ${SPACK_ROOT}/etc/spack/
+
+      - name: Checkout ${{ github.repository }}
+        uses: actions/checkout@v4
```
aidanheerdegen, line 110: IRL discussed, should be an input in case we ever need a dispatchable workflow

.github/workflows/ci.yml
```diff
@@ -0,0 +1,143 @@
...
+    container:
+      image: ghcr.io/access-nri/spack-ci:runner-rocky
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GHCR_PASSWORD }}  # FIXME: Set appropriate PAT
+      volumes:
+        - upstream:/opt/upstream:ro  # FIXME: Validate connection via container -> runner -> upstream pod
+    outputs:
+      spec-concretization-graph: ${{ steps.install.outputs.spec }}
+      # test-artifact-url: ${{ steps.test.outputs.artifact-url }}
```
aidanheerdegen, line 69: Would be good to have the lockfile output so users can recreate the CI run. Fine to be a link to an ephemeral artefact.

aidanheerdegen: Thought bubble: do we need a ROCrate object that stores more than just the lockfile, could also be packages, config, spack versions etc.

aidanheerdegen: Also return Sha of spack, spack-packages, spack-config and maybe the container Sha? And resolved git ref of branch.

aidanheerdegen (approved):
LSTM (Looks Superb To Me)

.github/workflows/ci.yml
CodeGat: I've realised that we would need to create a new output for each environment variable that we add to the potential action. I think that turning this into an action may actually lead to more lines of code that need to be changed to add a given environment variable (eg. add a variable as an action input, add an output in the action of the variable, etc. )

.github/workflows/ci.yml
CodeGat: Done in https://github.com/ACCESS-NRI/build-ci/pull/200/commits/669258e855e185cc9b637293f0309afdc1d3022b, https://github.com/ACCESS-NRI/build-ci/pull/200/commits/54e1e369f6826c13513ff16829a629762acbeba8

.github/workflows/ci.yml
CodeGat: Done in https://github.com/ACCESS-NRI/build-ci/pull/200/commits/669258e855e185cc9b637293f0309afdc1d3022b

CodeGat:
I've also added a `input.spack-compilers-manifest-path` that can contain compilers not installed upstream. This manifest will be installed before the actual testing one. See https://github.com/ACCESS-NRI/build-ci/pull/200/commits/f93c36df8b2148abe6b679f0b5b469ba78545f14

.github/workflows/ci.yml
aidanheerdegen: Fair enough.

Maybe just a loop
```suggestion
          for VAR in "SPACK_ROOT"  "SPACK_VERSION" "SPACK_CONFIG" "SPACK_PACKAGES"; do
              echo "$VAR=${!VAR}" >> $GITHUB_OUTPUT
          done
```

or to export all `SPACK_` related variables

```suggestion
          for VAR in ${!SPACK_*}; do
              echo "$VAR=${!VAR}" >> $GITHUB_OUTPUT
          done
```
(from https://stackoverflow.com/a/25765828)

.github/workflows/ci.yml
```diff
@@ -106,13 +145,33 @@ jobs:
...
+        with:
+          ref: ${{ inputs.ref }}
 
-      - name: Enable ci version of spack
-        run: . $SPACK_CONFIG_DIR/ci-spack-enable.bash
+      - name: Manifest - Install Compilers Locally
+        # Some compilers may not be available upstream, so we install them locally in a separate spack.yaml
+        if: inputs.spack-compiler-manifest-path != ''
+        run: |
+          spack env activate compilers --create --prompt --envfile ${{ inputs.spack-compiler-manifest-path }}
```
aidanheerdegen, line 161: This is cool! `spack` you magnificent bastard.

.github/workflows/ci.yml
```diff
@@ -106,13 +145,33 @@ jobs:
...
+          spack --debug install --fail-fast
+          spack env deactivate
+          yq '.spack.specs[]' ${{ inputs.spack-compiler-manifest-path }} | while read compiler; do
+            spack load $compiler
+          done
+          spack compiler find
+          spack compiler list
```
aidanheerdegen, line 168: This is really nice. I can see it becoming an action in the (your) future.

.github/workflows/ci.yml
```diff
@@ -106,13 +145,33 @@ jobs:
...
+
+      - name: Manifest - Replace usages of __REF__ with ${{ inputs.ref }}
+        # This means that the caller manifest files can remain unchanged, but still vary based on the ref
+        run: sed -i "s/__REF__/${{ inputs.ref }}/g" ${{ inputs.spack-manifest-path }}
```
aidanheerdegen, line 172: So the manifest is a template with `__REF__` replaced dynamically?

I think we might need to have a dialogue about future requirements. Should we standardise on a template approach, e.g. [jinja](https://jinja.palletsprojects.com/en/stable/)?

.github/workflows/ci.yml
```diff
@@ -124,6 +183,15 @@ jobs:
```
aidanheerdegen, line 183: I am confused. Is this embedding a HERETO doc in `$GITHUB_OUTPUT`?

aidanheerdegen (approved):
LGTM, but questions as always, and I think slipped in a suggestion separate from this review because wrong button.

.github/workflows/ci.yml
CodeGat: Yeah, it's a way of setting multiline output in github actions: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#example-of-a-multiline-string

.github/workflows/ci.yml
CodeGat: :eyes:

.github/workflows/ci.yml
CodeGat: Cool! Done in https://github.com/ACCESS-NRI/build-ci/pull/200/commits/457dbc5605d43a4fae96710d5e8bad51288711fd

.github/workflows/ci.yml
```diff
@@ -163,19 +170,32 @@ jobs:
...
-          spack compiler find
           spack compiler list
 
-      - name: Manifest - Replace usages of __REF__ with ${{ inputs.ref }}
-        # This means that the caller manifest files can remain unchanged, but still vary based on the ref
-        run: sed -i "s/__REF__/${{ inputs.ref }}/g" ${{ inputs.spack-manifest-path }}
+      - name: Manifest - Pip install jinja-cli for templating
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
```
aidanheerdegen, line 180: Unless there is a specific reason to use 3.13 I think we're standardising on a slightly older version, but the Python Standards Group would know.

.github/workflows/ci.yml
```diff
@@ -163,19 +170,32 @@ jobs:
...
+
+      - name: Manifest - Replace jinja template variables
+        # This means that the caller manifest files can remain unchanged, but still vary based on the {{ pr }}
+        # Additional data can be provided to fill in the jinja template through the `spack-manifest-data-path` input
+        id: jinja-templated
+        run: |
+          templated_manifest_path=$(dirname ${{ inputs.spack-manifest-path }})/$(basename ${{ inputs.spack-manifest-path }} .j2)
+          jinja \
+            -D pr ${{ inputs.ref }} \
+            ${{ inputs.spack-manifest-data-path != '' && format('-d {0}', inputs.spack-manifest-data-path) || '' }} \
```
aidanheerdegen, line 190: Is this missing the template file (`${{ inputs.spack-manifest-path }}`) to operate on?

aidanheerdegen (commented):
A couple of questions

.github/workflows/ci.yml
CodeGat: It was. I should have just copied the command after the test! Done in https://github.com/ACCESS-NRI/build-ci/pull/200/commits/57d6ed28dbbefe0902a48742b75943bd26b3fce3

.github/workflows/ci.yml
CodeGat: `3.11` it is! Done in https://github.com/ACCESS-NRI/build-ci/pull/200/commits/57d6ed28dbbefe0902a48742b75943bd26b3fce3

aidanheerdegen (approved):
LGTM!

PR #207, opened 2025-05-06
Update models.json

### LINKED ISSUE #201: Update components for access-om3
When [om3 component spack packages PR](https://github.com/ACCESS-NRI/spack-packages/pull/171) is merged

We need to add the final names here:

https://github.com/ACCESS-NRI/build-ci/blob/461a98f2b109ea6f2a974106648b3699258b1ce2/config/models.json#L3

I assume these are the names for the spack-packages?

anton-seaice:
@CodeGat - is this change correct?

What happens when more than more than one model uses the same component ?

closes #201

### DIFF
--- a/config/models.json
+++ b/config/models.json
@@ -1,6 +1,6 @@
 {
     "access-om2": ["libaccessom2", "mom5", "cice5", "oasis3-mct"],
-    "access-om3": ["access-om3"],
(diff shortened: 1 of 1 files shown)


CodeGat (approved):
Yep, that looks good. More than one model is allowed to use the same component, it's just used to figure out what components are part of a given model.

PR #210, opened 2025-06-17
Output SHAs even when refs aren't updated

CodeGat:
## Background

See https://github.com/ACCESS-NRI/MOM5/actions/runs/15695675683/job/44220448187#step:2:143

When we use the default values for `*-ref` inputs, we skip the relevant repo update steps. However, this is how we get the SHAs that are outputted, leading to no SHAs outputted for repos that use defaults. 

In this PR, we no longer skip the update action steps, and output the SHA (updated or not) in that action. We also add a new output `updated` that allows a later relinking step (between `spack` and `spack-config`) only go ahead if those repos were updated. 

## The PR

In this PR:

- **Update git-checkout-updated-ref to note whether the ref was actually updated**
- **Always attempt to update refs, and only relink config to spack if actually updated**
- **Delete util folder as we don't need buildcache keys**
- **TEST: Update action ref for testing**

## Testing

Tested on `MOM5` here: https://github.com/ACCESS-NRI/MOM5/actions/runs/15698218709/job/44227413029 using all defaults (shows how quickly things can go out of date!)


tmcadam (approved):
Reviewed with Tommy. Looks good.

PR #212, opened 2025-06-19
README Update

CodeGat:
Closes #211

## Background

This is a big change in infrastructure, so we should update the documentation, as well!

## This PR

- **Remove README-DEV.md**
- **Update README.md for build-ci 2.0**
- **Update containers README.md**

Issue #193, opened 2025-02-17
Consider Post-Build Testing Requirements


CodeGat:
A probable output to the Entrypoint workflows would be something like a link to a singularity container with the given model component installed. But model components need more testing than that, and this testing could be quite varied between model components. 

## Things to Consider

* Would there be a common core of tests that need to be run? For example, testing if the model component can be installed as a dependent to multiple different models. 
* Should we just return the built image, in which case:
  * Would we return a reference to a singularity image with the model installed? Similar to a `[docker commit](https://stackoverflow.com/questions/62698416/what-is-the-equivalent-of-docker-commit-in-singularity)` image? 
  * Or a reference to a running container on GHA? How would this work for private repository minutes?

CodeGat:
Rather than a pytest wrapper for commonality, maybe (also?) support running post-build scripts under `.github/build-ci/test-scripts`, or something, that is executed in the container. Might be easier for model devs rather than having to call pytest to call a script.

aidanheerdegen:
An example of a post-build test would be that the `pkgconfig` file has been correctly generated (if there is one), e.g.


https://github.com/ACCESS-NRI/access-spack-packages/blob/main/packages/oasis3-mct/package.py#L78-L91

Issue #165, opened 2024-06-17
Read `target` from `config/compilers.json`


CodeGat:
Hi @CodeGat , Let me know when the PR to read `target` from `compilers.json`, is ready for review.

_Originally posted by @harshula in https://github.com/ACCESS-NRI/build-ci/pull/164#discussion_r1642046627_

Issue #158, opened 2024-05-10
Can we have images that use the `UM` be `internal` or `private`, rather than `public`?


CodeGat:
Since `UM` related builds cannot be public, we cannot at this stage build `access-esm1.5` using `build-ci`. 

Investigate the possibility of making these docker images `private` or `internal`, and note what kind of credentials one would need to access it as part of a run. 

See https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility#ensuring-workflow-access-to-your-package

CodeGat:
Notes: Indeed we can make the packages `internal` (viewable to members of the org) or `private` (only to people who have access), but it seems that we cannot do this programmatically, only via the GUI.
![image](https://github.com/ACCESS-NRI/build-ci/assets/45781416/c957a3c5-799f-484c-b3d8-40714f3c1e90)
And even if we can make it private:
> If you grant a public repository access to private packages, forks of the repository may be able to access the private packages.

CodeGat:
Trying a different tack. We will now no longer automatically link the visibility of the package to the access rights of the repository on an organisational level. See https://github.com/organizations/ACCESS-NRI/settings/packages

CodeGat:
This hasn't seemed to have helped the visibility of the package itself. See https://github.com/ACCESS-NRI/build-ci/pkgs/container/build-cable-intel2021.2.0-2024.06.05

CodeGat:
Closing due to [`build-ci 2.0`](https://github.com/ACCESS-NRI/build-ci/pull/213) update

PR #213, opened 2025-06-25
Build CI 2.0

CodeGat:
Closes project https://github.com/orgs/ACCESS-NRI/projects/30, but remaining issues will be incorporated into later projects. 

References issues now not planned due to `2.0` update: #165, #158, #151, #128, #119, #112, #106, #105, #90, #72, #48, #8, ACCESS-NRI/build-cd#2, ACCESS-NRI/build-cd#1.

> [!IMPORTANT]
> This update requires changes to both the default branch and the set of allowlisted refs for the runner group

## Background

This PR updates the `build-ci` pipeline to be faster, more customisable, and less flaky. It's so different, in fact, I called it `build-ci 2.0`. Some of the main features include:

- **Spack manifest recipes**: Ability to specify what is installed in spack manifest format - this allows customisation of compilers, flags, versions and configuration where there was none before. You can install the model component as part of a larger model, with different flags and config, with no restriction on the style of the manifest...which leads into the next feature...
- **Manifests Installable in Parallel**: The `ci.yml` entrypoint is designed to take advantage of GitHub Actions own `strategy.matrix`, so you can build and test multiple models that a given component is part of, at the same time. 
- **More Customisable Stuff**: Not only are the manifests more customizable, but the versions of `spack-packages`, `spack-config` and `spack` itself are all customisable at the PR level. 
- **SSH Into The Installation**: Debug further once the install completes (or fails!) and check out other logs. Only the PR Author can do this.

This was more than just an update to this repo, though. It also led to the standing up of a Nectar Kubernetes cluster for ephemeral self-hosted GitHub Runners, which can be found in `ACCESS-NRI/build-ci-k8s-infra` (private repository). 

This update to `build-ci` also puts it in good stead to be updated into the future, with [unit testing](https://github.com/ACCESS-NRI/build-ci/issues/193) and other optimisations coming later. 

## This PR

The following PRs are reviewed and incorporated into this this PR:

- #196
- #200
- #203
- #210
- #212

Fixes on branch have also been incorporated directly into this PR as tests have been run - more details can be found in the commit list. 

## Testing

Testing details for each Pull Request can be found in the respective Pull Request. 

### Kubernetes-independent Testing

Tests of the `containers` directory included ability to link to the upstream installation, ability to install and use compilers from the runner, ability to install private packages, and use of the buildcache. 

Similar tests can be run locally via:

```bash
COMPOSE_BAKE=1 docker compose -f containers/compose.dev.yml build
docker compose -f containers/compose.dev.yml up --detach
docker compose -f containers/compose.dev.yml exec runner bash
# and once finished...
docker compose -f containers/compose.dev.yml down --volumes
```

### Kubernetes Architecture Testing

In terms of the overall solution, the architecture underwent a stress test (runs found here: https://github.com/ACCESS-NRI/MOM5/pull/48). It was noted that we woulds still be well under our CPU/memory/storage limits imposed by Kubernetes if we set the maximum amount of runners to 8. This is open to change as required. 

This balancing act of maxRunners is required as Kubernetes kills pods that are exerting DiskPressure/MemoryPressure/etc on the Node, which would lead to runners being destroyed mid job, which is immensely unsatisfying for the user. Setting maxRunners also means that jobs will wait for free runners.  

## More information

More information on the infrastructure and design decisions can be found in the private `build-ci-k8s-infra` repository at https://github.com/ACCESS-NRI/build-ci-k8s-infra, as well as the Decision Document here: https://github.com/ACCESS-NRI/dev-docs/wiki/Decision-Documents#build-ci-20


tmcadam (approved):
Had another run through of this with Tommy on Friday, looked good.

CodeGat:
Thanks @tmcadam, I'll await @aidanheerdegen s review before merging :smile:

.github/workflows/ci.yml
```diff
@@ -189,7 +192,7 @@ jobs:
         # Additional data can be provided to fill in the jinja template through the `spack-manifest-data-path` input
         id: jinja-templated
         run: |
-          templated_manifest_path=$(dirname ${{ inputs.spack-manifest-path }})/$(basename ${{ inputs.spack-manifest-path }} .j2)
+          templated_manifest_path=$(dirname ${{ inputs.spack-manifest-path }})/templated.$(basename ${{ inputs.spack-manifest-path }})
```
aidanheerdegen, line 195: I am an opinion-haver, and there they are:

I think it is a bit confusing for users that the templated manifests have changed their names by having `templated.` prefixed to them.

Suggested options:

1. Insist on a `j2` suffix and then drop that for the templated manifest. This looks nicer, but requires more code to enforce.

Or

2. Save them to a `templated` subdir with the `j2` stripped off (if it exists, otherwise do nothing)

.github/workflows/ci.yml
```diff
@@ -203,6 +206,10 @@ jobs:
 
       - name: Manifest - Install
         id: install
+        env:
```
aidanheerdegen, line 209: Can you just add a little comment to explain what this bit does?

README.md
```diff
@@ -2,49 +2,85 @@
...
 
-## Usage
+### Writing Spack Manifests
+
+Spack [manifest files](https://spack.readthedocs.io/en/latest/environments.html) are a way to define versions, dependencies (and dependents) and configuration information for software. In this case - model components, and what models they are part of.
+
+The manifest files have almost no restrictions - but in order to build the component at the git ref given by the workflow input `inputs.ref` (which for pull requests is the HEAD of the source branch), one must use the version `@git.{{ ref }}` for the model component.
 
-### For Model repositories
+The double-bracket syntax is unique and not spack specific - the manifest files are actually Jinja templates, in which `{{ ref }}` is replaced by `inputs.ref` at build-time. Additional template variables can be defined through data files that are ingested via the `inputs.spack-manifest-data-path` entrypoint workflow input. You can add the `.j2` suffix to the spack manifest file name, but it isn't required.
```
aidanheerdegen, line 58:
```suggestion
The double-bracket syntax is unique and not spack specific - the manifest files are actually [Jinja](https://palletsprojects.com/projects/jinja/) templates, in which `{{ ref }}` is replaced by `inputs.ref` at build-time. Additional template variables can be defined through data files that are ingested via the `inputs.spack-manifest-data-path` entrypoint workflow input. You can add the `.j2` suffix to the spack manifest file name, but it isn't required.
```

containers/Dockerfile
```diff
@@ -221,29 +222,10 @@ rm -rf static-tmate.tar.xz
 EOF
 
 # Set up ACCESS Spack buildcache
-RUN pip3 install --no-cache-dir boto3==1.23.10
-RUN spack gpg init
+RUN spack mirror add --autopush --unsigned runner_set_buildcache /opt/runner_set_buildcache
```
aidanheerdegen, line 225: So cool to get that level of functionality with a single simple command. Nice one spack!

containers/README.md
```diff
@@ -1,25 +1,29 @@
...
+Furthermore, those images download upstream compilers and common packages informed by the `upstream/[dev|prod]/[packages|compilers].spack.yaml` spack manifests.
 
-    docker build -f Dockerfile.base-spack -t <name>:<version> --target dev --no-cache --progress=plain .
+Finally, there is a `spack-config` folder that contains a customised [spack-enable.bash](https://github.com/ACCESS-NRI/spack-config/blob/main/spack-enable.bash) used to load `spack`, as well as upstream compilers.
 
-To build an image with specific versions of `spack`, `spack-packages` and `spack-config`:
+## How to build `spack` for testing
 
-    docker build -f Dockerfile.base-spack -t spack-dev:v0.20 --target dev --build-arg SPACK_VERSION=v0.20 --build-arg SPACK_PACKAGES_REPO_VERSION=2024.03.22 --build-arg SPACK_CONFIG_REPO_VERSION=2024.03.22 --no-cache --progress=plain .
+You can spack in a similar way to `build-ci`, with it's own upstream `spack` used for compilers and common packages. This is also suitable for developers to test. Run the following Docker Compose commands:
```
aidanheerdegen, line 15:
```suggestion
You can use spack in a similar way to `build-ci`, with it's own upstream `spack` used for compilers and common packages. This is also suitable for developers to test. Run the following Docker Compose commands:
```

.github/workflows/README.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+#### Simple Matrix
+
+```yaml
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      max-parallel: 2
```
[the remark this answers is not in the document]
CodeGat: Good point. Will add.

containers/Dockerfile
CodeGat: It is surprisingly easy. Same command for that `build-cd` buildcache, as well

.github/workflows/ci.yml
CodeGat: Done in https://github.com/ACCESS-NRI/build-ci/pull/213/commits/969a9db9b19fc0d814b6623a4e09cc17493fab72

.github/workflows/ci.yml
CodeGat: Fair enough, I went with option 2 in https://github.com/ACCESS-NRI/build-ci/pull/213/commits/fe827574b518a154e2c9e964a1cdf419116a55b7 as I want to give the user FREEDOM!!

aidanheerdegen (approved):
Awesome! Looks great, nice work.

aidanheerdegen:
🚀 🥳 🎈

PR #217, opened 2025-07-18
Build-ci 1.0 - Checkout `v1` branch for config rather than default branch

CodeGat:
For repositories still using the earlier version of `build-ci`, they checkout configuration information from the default branch. Since later versions of `build-ci` don't use this configuration information, the workflow fails. 

## The PR

* Pin the ref that is checked out for the earlier version of `build-ci`  to the `v1` branch


aidanheerdegen (approved):
LGTM

PR #218, opened 2025-07-21
Update Upstream Image Packages and Dependencies

CodeGat:
## Background

CABLE-LSM/CABLE is using a prototype GitHub-hosted CI 2.0 without a buildcache at the moment. This means that builds are actually slower than the original, everything-except-the-dependent way. 

To get it as close as we can, we should include relevant versions of `netcdf-c` and `netcdf-fortran` in the upstream image. 

Furthermore, since we can't share data between service containers in GitHub-hosted mode, we have elected to just run the upstream container as the runner container. This means that the upstream image requires tmate to be installed. 

## The PR

- **Add tmate to both runner and upstream images**
- **Add netcdf-c@4.9.2 and netcdf-fortran@4.5.2 to upstream packages**


CodeGat:
1. Yep!
2. Currently building and will let you know when it's done :)

PR #219, opened 2025-07-22
GitHub-hosted Build CI 2.0

### LINKED ISSUE #216: Allow non-Self-Hosted Runners to Run `build-ci 2.0`
There is a potential use case for non-self-hosted runners to use this infrastructure. 

Technically, the only thing that we would need to provide to run the container is a R/W volume to a buildcache that is somehow on the runner, and a R volume to an upstream instance of spack that is also on the runner. 

This could be done via a `jobs.*.runs-on.groups: [build-ci, build-ci-github]` directive, with each being mutually exclusive. 

One issue arises with how do we connect up the volumes to external sources within the same `container` job? Dependency jobs can't affect future jobs in this way, and we can't affect stuff out of the container in the jobs itself. Hmmm...

CodeGat:
Closes #216

## Background

There are some use cases for `build-ci 2.0` outside of a self-hosted runner - for example, if we wanted external orgs to use it. 

In this PR, we have migrated the original workflow to a self-hosted format, under `ci-github-hosted.yml`. 

>[!IMPORTANT]
> There are some caveats on using GitHub-hosted over self-hosted - currently there is no buildcache, and we use the upstream image as the runner image - leading to a larger (1 min 30 sec) setup phase due to downloading of a large image. 

## The PR

* Create a `ci-github-hosted.yml` that allows use of Build CI 2.0 outside of the organisation.`

## Testing

Testing was done primarily on https://github.com/CABLE-LSM/CABLE/pull/600 - validated the workflow works fine, is similar in time to CI 1.0, and can SSH into the install as required.


CodeGat:
Note that I know there is a lot of duplication in this added file, but it can't be helped until we have a solution to the volumes section of the workflow, which is the only section that must be different between the two workflows - everything else is able to be merged together.  See #220.

aidanheerdegen (approved):
LGTM

PR #235, opened 2025-08-12
Relink `spack-config` -> Updated `spack` with appropriate config dir

### LINKED ISSUE #234: Relinking `spack-config` uses the `SPACK_CONFIG_DIR` from container
When we update spack to a new major in `build-ci`, it relinks the `spack-config`. However, it uses `0.22` by default as it uses the environment variable `SPACK_CONFIG_DIR` which is set at build time. 

Recompute this value when we update to a new major version of `spack`.

CodeGat:
Closes #234

## Background

Users have tried updating their `spack` to `1.0`, but to relink `spack-config` to `spack` we always use the value of `SPACK_CONFIG_DIR`, which is set at build time. This means that we can attempt to link `v0.22` config to `v1.0`, which is bad. 

## The PR

This PR checks the ref that `spack` was updated to, then determines if that is part of a given `origin/releases/vX` branch, and then relinks `spack` to `spack-config` via the appropriate `vX` directory.

## Testing

Tested via https://github.com/ACCESS-NRI/access-test-component/pull/9

### Test default spack linking

Succeeded in run https://github.com/ACCESS-NRI/access-test-component/actions/runs/16922975520/job/47953036403#step:7:20

### Test `HEAD` of `releases/v0.23` linking

Succeeded in run https://github.com/ACCESS-NRI/access-test-component/actions/runs/16923099192/job/47953409130#step:7:20 - cancelled rest of the run to save time

### Test commit within `releases/v1.0` linking

Succeeded in run https://github.com/ACCESS-NRI/access-test-component/actions/runs/16923169567/job/47953615479#step:7:20 - rest of run failed due to wrong `spack-packages` version used

### DIFF
--- a/.github/workflows/ci-github-hosted.yml
+++ b/.github/workflows/ci-github-hosted.yml
@@ -205,7 +205,7 @@ jobs:
         # Furthermore, if spack has been updated, we need to make sure we link to the correct vX config directory in spack-config.
         # We find what branch the spack SHA is part of, and attempt to link to that directory.
         run: |
-          spack_branch=$(git -C ${{ steps.env.outputs.SPACK_ROOT }}/../spack for-each-ref --format='%(refname:short)' refs/remotes --contains ${{ steps.spack-update.outputs.sha }} \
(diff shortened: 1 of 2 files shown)


aidanheerdegen (approved):
ngl I am a bit worried about the fragility of the ref/branch inspection, but if it breaks in the future we'll just have to fix it, and the consequences are pretty low stakes as it won't break deployments etc.

PR #244, opened 2025-09-04
Scope both the local and OCI buildcaches to `system` rather than `user`

### LINKED ISSUE #243: Self-Hosted: Issues with Local Buildcache
Currently there is an issue with self-hosted local buildcaches where the `user`-scoped buildcache exists during the build phase, but is inaccessible during the running of the container. 

When investigating via `ssh`ing into the running container (as part of an `access-test-component` run, I noted that although the `/root/.spack/mirrors.yaml` existed, it was not readable via `spack config blame mirrors`. 

When running the container locally, the config is as it should be:

```
[root@63bbd78dd3ad /]# spack config blame mirrors
---                                           mirrors:
/root/.spack/mirrors.yaml:2                     runner_set_buildcache:
/root/.spack/mirrors.yaml:3                       url: /opt/runner_set_buildcache
/root/.spack/mirrors.yaml:4                       autopush: True
/root/.spack/mirrors.yaml:5                       signed: False
/opt/spack/etc/spack/defaults/mirrors.yaml:2    spack-public:
/opt/spack/etc/spack/defaults/mirrors.yaml:3      binary: False
/opt/spack/etc/spack/defaults/mirrors.yaml:4      url: https://mirror.spack.io

```

Very weird. Attempt to move the initial `runner_set_buildcache` to the `system` scope, and see if it is visible.

CodeGat:
Closes #243

## Background

For some reason `--scope=user` buildcaches are not persisted when running a container as `root` via GitHub Actions, even though they work locally in the container. 

The fix is to move the `spack mirror add` to the `--scope=system` level, rather than the `user` one. 

## The PR

* Update the scope while building the image to `system`, from `user`
* Also update the `oci`-backed buildcache scope to `system`, just to be consistent. 

## Testing

Tested via https://github.com/ACCESS-NRI/access-test-component/actions/runs/17453892656/job/49563693372?pr=12#step:14:1358 (in which you'll note the searching for buildcache items)

### DIFF
--- a/.github/workflows/ci-github-hosted.yml
+++ b/.github/workflows/ci-github-hosted.yml
@@ -273,7 +273,8 @@ jobs:
         if: inputs.spack-oci-buildcache-url != ''
         run: |
           . ${{ steps.env.outputs.SPACK_ROOT }}/share/spack/setup-env.sh
-          spack mirror add --unsigned \
+          spack mirror add --scope=system \
+            --unsigned \
             --oci-username-variable OCI_USERNAME \
             --oci-password-variable OCI_PASSWORD \
             oci_buildcache ${{ inputs.spack-oci-buildcache-url }}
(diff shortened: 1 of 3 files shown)


aidanheerdegen (approved):
LGTM

PR #245, opened 2025-09-05
Use Hashes for Spec Logs Generation

CodeGat:
See https://github.com/ACCESS-NRI/access3-share/actions/runs/17455915051/job/49569374009, from https://github.com/ACCESS-NRI/access3-share/pull/11

## Background

Using specs for log generation has the possibility of failing due to spec definitions potentially being quite similar (see https://github.com/ACCESS-NRI/access3-share/pull/11/files#diff-9664d220045bd40cfb26c542411ad41aeab22c0d1334418d10d67757fdbd26bfR3-R4)

Instead of using the spec name, use the hash instead, which is unambiguous. 

## The PR

* `ci[-github-hosted].yml`: Replace generation of logs via spec name with spec hash

## Testing

Tested in `ghcr.io/access-nri/build-ci-upstream:rocky-2025.09.001`, by running https://github.com/ACCESS-NRI/access3-share/pull/11/files#diff-9664d220045bd40cfb26c542411ad41aeab22c0d1334418d10d67757fdbd26bf and then running:

https://github.com/ACCESS-NRI/build-ci/blob/1fe40b88f8821ce644174ae2afec64fa256a6015/.github/workflows/ci.yml#L392-L400 (subsituting the GitHub Actions Templating).

### DIFF
--- a/.github/workflows/ci-github-hosted.yml
+++ b/.github/workflows/ci-github-hosted.yml
@@ -384,16 +384,16 @@ jobs:
             exit 0
           fi
 
-          # Get all the top-level specs from the manifest
-          specs="$(jq --raw-output '.roots[].spec' ${{ steps.env.outputs.spack-env-dir }}/default/spack.lock)"
-          echo "Found top-level specs:"
-          echo "$specs"
+          # Get all the top-level hashes from the manifest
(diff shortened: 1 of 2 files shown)


aidanheerdegen (approved):
LGTM

PR #246, opened 2025-09-11
Load compilers via upstream lockfile

CodeGat:
## Background

Currently, we load compilers via the `compilers.spack.yaml` file. Since this is abstract, there is a possibility that the spec doesn't align with later/different versions of the upstream. Hashes (via a lockfile) are unique and persistent. 

## The PR

* Use the `compilers.spack.lock` file stored in the upstream volume to unambigiously refer to the compiler packages
* Also bootstrap spack as part of the runner image, so we don't take 1 minute to bootstrap every time we spawn a runner

## Testing

Done in https://github.com/ACCESS-NRI/MOM6/actions/runs/17629880394/job/50095056910


jo-basevi (approved):
LGTM

Issue #247, opened 2025-09-19
`1.0` Runners Using `0.22` Upstream Can't Ascertain Compiler For Upstream Packages


CodeGat:
## Background

Currently, we are running a `0.22` Upstream image. This was chosen as `0.22` Runners are not able to read `1.0` Upstreams. 

However, we now know that `1.0` Runners (while they work) can't ascertain the correct compiler for packages in the `0.22` upstream, leading to almost no packages being reused from the upstream.

## Implementation Plan

At some point, we will have to do a direct cutover to `1.x` similar to what we have to do for `build-cd`, in which:

* We replace the `0.22` Upstream with a `1.x` one.
* We enforce a `>=1.0` Runner for all model component repositories
* Which means we need to update all manifests to use language constraints (like `spack.packages.c.require: [COMPILER_PACKAGE@VERSION]` over `spack.packages.all.require: [%COMPILER@VERSION]`

CodeGat:
Currently testing out a solution in which we have two upstreams on the kubernetes cluster - one for `0.22` and one for `1.0`, which are both hooked into the runner (read only). The correct upstream for the given spack version on the runner is picked via spack-version-specific `spack-config`. 

Currently we can source the correct upstream via spack-config, but the `config.install_tree` in the upstream (`/opt/upstream`) does not match the `upstreams.upstream.install_tree` mounted volume (`/opt/upstream-v1.0`). 

Currently testing rebuilding the `v1.0` image with the `config.install_tree` being the same as the runners `upstreams.upstream.install_tree` mounted volume (`/opt/upstream-v1.0`)

CodeGat:
It works!

CodeGat:
This is complete via the merged linked PRs

PR #248, opened 2025-09-22
Support multiple upstream volumes with different spack versions

CodeGat:
References #247

## Background

Currently, for `0.22` and `1.0` spack runners, we use a `0.22` upstream image for compatibility between the two. 

Unfortunately, the `1.0` runner doesn't understand what compilers are used in a `0.22` upstream, noting them as `no compiler`. This means that there is no upstream caching for `1.0` images. 

The solution that has been decided is to have two upstream volumes (`0.22` and `1.0`) in the cluster (see PR https://github.com/ACCESS-NRI/build-ci-k8s-infra/pull/3), and the determination for which to actually consider an upstream to be left to `spack-config` (see PR https://github.com/ACCESS-NRI/spack-config/pull/78).

## The PR

- **Add two read-only volumes for both 0.22 and 1.0 upstreams**
- **Load specific compilers.spack.yaml from chosen upstream**

## Testing

Done in https://github.com/ACCESS-NRI/MOM6/pull/30 (see run https://github.com/ACCESS-NRI/MOM6/actions/runs/17904675462/job/50903910096?pr=30#step:8:24, noting that the workflow failure is due to a perl bug that is fixed with a separate `spack-config-ref`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,7 +175,7 @@ jobs:
         # FIXME: To be able to support both v0.22 and v1.0 upstreams at the same time
         # We need to pick the correct upstream volume based on the runner image spack version
         # This may not be required in the future if we only support v1.0+
-        - /opt/upstream${{ contains('v1.0', inputs.container-image-version) && '-v1.0' || '' }}:/opt/upstream${{ contains('v1.0', inputs.container-image-version) && '-v1.0' || '' }}:ro
+        - /opt/upstream${{ contains(inputs.container-image-version, 'v1.0') && '-v1.0' || '' }}:/opt/upstream${{ contains(inputs.container-image-version, 'v1.0') && '-v1.0' || '' }}:ro
         - /opt/runner_set_buildcache:/opt/runner_set_buildcache:rw
     outputs:
       spec-concretization-graph: ${{ steps.install.outputs.spec }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -169,7 +169,6 @@ jobs:
     runs-on:
       group: build-ci
     container:
-      # image: ghcr.io/access-nri/build-ci-runner:${{ inputs.container-image-distro }}-${{ inputs.spack-version }}-${{ container-image-version }}
       image: ghcr.io/access-nri/build-ci-runner${{ inputs.container-image-version }}
       volumes:
         # FIXME: To be able to support both v0.22 and v1.0 upstreams at the same time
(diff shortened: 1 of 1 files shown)


aidanheerdegen (dismissed):
LGTM

CodeGat:
Updated the PR to instead find the correct upstream volume based on the `inputs.container-image-version` (despite it being the runner) rather than via `spack-config`. See https://github.com/ACCESS-NRI/build-ci/pull/248/commits/3ab71d488ca6fccdf6ea131a9b092a5b066ad012. Updated testing done in https://github.com/ACCESS-NRI/MOM6/actions/runs/17937970010/job/51007842730?pr=30

.github/workflows/ci.yml
```diff
@@ -169,9 +169,14 @@ jobs:
...
     container:
+      # image: ghcr.io/access-nri/build-ci-runner:${{ inputs.container-image-distro }}-${{ inputs.spack-version }}-${{ container-image-version }}
       image: ghcr.io/access-nri/build-ci-runner${{ inputs.container-image-version }}
       volumes:
-        - /opt/upstream:/opt/upstream:ro
+        # FIXME: To be able to support both v0.22 and v1.0 upstreams at the same time
+        # We would enable each upstream in the spack-config, depending on the spack version
+        # This may not be required in the future if we only support v1.0+
+        # TODO: Update image tags to include vX spack versionL eg. :rocky-v0.22-2025.09.003
+        - /opt/upstream${{ contains('v1.0', inputs.container-image-version) && '-v1.0' || '' }}:/opt/upstream${{ contains('v1.0', inputs.container-image-version) && '-v1.0' || '' }}:ro
```
aidanheerdegen: Not a huge fan of this due to readability and fragility issues, but I will hold my nose and approve.

aidanheerdegen (dismissed):

.github/workflows/ci.yml
CodeGat: I'm not a fan of it either. But outside of the original solution of letting `spack-config` pick upstreams based on the upstream version, or adding an essentially duplicate input for the branch `spack` is on, this is the only solution given some of the constraints on GitHub Actions syntax.

.github/workflows/ci.yml
```diff
@@ -169,9 +169,13 @@ jobs:
     runs-on:
       group: build-ci
     container:
+      # image: ghcr.io/access-nri/build-ci-runner:${{ inputs.container-image-distro }}-${{ inputs.spack-version }}-${{ container-image-version }}
```
aidanheerdegen: Do you want to leave this commented line in?

aidanheerdegen (dismissed):
Now that seems a lot cleaner. I likeee!

.github/workflows/ci.yml
CodeGat: WHOOPS

aidanheerdegen (approved):

PR #252, opened 2025-10-03
GitHub-Hosted: Re-find own compilers

### LINKED ISSUE #251: GitHub-hosted: Doesn't treat own compilers as externals, needs to be re-found
See https://github.com/CABLE-LSM/CABLE/actions/runs/18214132941/job/51859978903?pr=618 among others around that date

## Background

For some reason, github-hosted runs of `build-ci` also seem to fail to find compilers when their version of spack has been updated. Therefore, it will need a similar step of finding their own compilers from their own "upstream"

CodeGat:
Closes #251

## Background

For some reason, updating to later versions of spack made it forget that the compilers were external. This step attempts to refind them from either `/opt/upstream/compilers.spack.lock` (for runners, implemented previously) or `/opt/environments/compilers/spack.lock` (for upstream, this PR). 

## The PR

- **Find compiler lockfile either from the upstream volume (if it exists) or the compilers environment**
- **Change local compiler environment name so it doesn't clash**

## Testing

Testing done in https://github.com/CABLE-LSM/CABLE/pull/618, specifically https://github.com/CABLE-LSM/CABLE/actions/runs/18214875186?pr=618

### DIFF
--- a/.github/workflows/ci-github-hosted.yml
+++ b/.github/workflows/ci-github-hosted.yml
@@ -263,17 +263,23 @@ jobs:
         run: |
           . ${{ steps.env.outputs.SPACK_ROOT }}/share/spack/setup-env.sh
 
-          # Attempt a load of compilers via upstream compilers lockfile (if it exists)
-          if [ -f /opt/upstream/compilers.spack.lock ]; then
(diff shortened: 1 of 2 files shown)


aidanheerdegen (approved):
Worth a crack on a Friday arvo

Issue #220, opened 2025-07-22
Merge `ci.yml` and `ci-github-hosted.yml` Once Solution to Volumes Found


CodeGat:
## Background

Currently we require two separate files to handle self-hosted and GitHub-hosted runs of the pipeline. They are functionally the same, and is thus a lot of duplication. 

## Implementation Plan

Currently, progress is being help up by the `volumes` section, which is either two entries if self-hosted, and none if github-hosted. 

The ways that this can be solved are:

* Either we need to be able to have the GitHub-hosted runs accept the two volumes, which would be essentially empty or useless, or
* We can put two-volumes-for-self-hosted, none-for-github-hosted in an expression that Actions finds valid, or
* GitHub-hosted runs allow sharing of data via volumes between a main container and a service container, in which the main container is the runner variant of the image, and the service container is the upstream variant of the image
Current progress can be found in ACCESS-NRI/build-ci@48c9ed01e02d773feefe0db502c3bf4a5fdbc208

CodeGat:
Opened a ticket regarding this issue

PR #257, opened 2025-11-06
Output `spack config blame`, Continue on error for Log Creation

### LINKED ISSUE #256: Output config before installation
## Background

Config is currently a bit of a black box, unless one inspects it manually via `inputs.allow-ssh-into-spack-install` and doing it themselves. 

## Implementation details

* Output the `spack config blame` before `spack install`

CodeGat:
Closes #256

## Background

Currently the configuration used in an install is quite a black box, only inspectable if one has `inputs.allow-ssh-into-spack-install`

## The PR

* Output `spack config blame` in a `Spack Config Blame Output` group
* Also don't error out if logs can't be created, as this may be due to an earlier install failure, which we should point users towards instead.

### DIFF
--- a/.github/workflows/ci-github-hosted.yml
+++ b/.github/workflows/ci-github-hosted.yml
@@ -362,6 +362,9 @@ jobs:
           echo "$(spack spec)" >> $GITHUB_OUTPUT
           echo "EOF" >> $GITHUB_OUTPUT
 
+          # Output config before install for debugging
+          spack config blame
+
           spack --debug install --fail-fast --deprecated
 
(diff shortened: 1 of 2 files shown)


harshula:
Hi @CodeGat , Is there a way to avoid duplicating code in both files?

CodeGat:
Unfortunately not - see https://github.com/ACCESS-NRI/build-ci/issues/220. I've been in contact with GitHub to resolve the underlying issue, which I hope they can get back to me soon.

harshula (approved):
🤞

PR #261, opened 2025-12-09
Merge `ci-github-hosted.yml` into `ci.yml` using new, optional `inputs.run-self-hosted`

### LINKED ISSUE #220: Merge `ci.yml` and `ci-github-hosted.yml` Once Solution to Volumes Found

CodeGat:
Closes #220

## Background

Currently, `build-ci` uses two workflow files, `ci.yml` and `ci-github-hosted.yml` for self- and GitHub-hosted runners respectively. 

This is due to small, previously irreconcilable differences in the `jobs.*.containers` section - specifically regarding the `volumes` section. The GitHub-hosted variant didn't need any volumes to function, but the self-hosted one did.

After chatting with GitHub Support, it was noted that it is indeed possible to reconcile these differences, by using a `inputs.run-self-hosted` input that turns the volume list into the expected list of volume defs if `true`, and empty lists if `false`. 

## The PR

* Add new, optional `inputs.run-self-hosted` that defaults to `true`
* `jobs.*.runs-on` is determined by `inputs.run-self-hosted` - either the self-hosted group or the default `ubuntu-latest`
* `jobs.*.container.image` is also determined by the new input - picking either the runner or upstream image
* `jobs.*.container.volumes` is the same - either the two required volumes for self-hosted, or dummy volumes for GitHub-hosted

## Testing

Tested in https://github.com/ACCESS-NRI/access-test-component/pull/15. Note that the failures are due to concretization issues with the manifest - it is the initialisation of the containers that is the important part. 

Specifically:

* GitHub-hosted pulled the correct upstream image with dummy volumes in https://github.com/ACCESS-NRI/access-test-component/actions/runs/20053240115/job/57513298767 (see Initialise Containers step)
* Self-hosted pulled the correct runner image with two volumes in https://github.com/ACCESS-NRI/access-test-component/actions/runs/20051632302/job/57508743153 (see Initialise Containers step)

## Other Implementations

In https://github.com/ACCESS-NRI/build-ci/pull/261/commits/f2c8bd2a93e8043e9570557882c328d2efd9ce1c, I used `fromJson('[]')` as the GitHub-hosted case, but that was a bit flaky, as it was doing:
```yaml
- ${{ fromJson('[]') }}
- "[]"
``` 
Which meant under the hood `docker --volume ` was silently ignoring the invalid volume definition (usually of the form `HOST_PATH:CONTAINER_PATH:FLAGS`). This works okay, but is flaky - docker or GitHub may throw errors for that in future.


aidanheerdegen (approved):
Nice work!

PR #262, opened 2025-12-11
Make default `builtin-spack-packages-ref` from `repos.yaml`

CodeGat:
References https://github.com/ACCESS-NRI/access-spack-packages/pull/353#discussion_r2605316172

## Background

The current default for `inputs.builtin-spack-packages-ref` is the branch `develop`. This is consistent with the other `inputs.*-ref` inputs. However, we are overwriting the currently-set `spack-config` `repos.yaml`.

Instead of the default being the bleeding edge of `spack/spack-packages`, we should default to the ref in the `repos.yaml` in `spack-config` at the `inputs.spack-config-ref`. If there is no value in there either, it should be `develop` (as is `spack`s default. 

This would mirror the Dockerfile in which `spack-config` informs the initial versions of the spack packages repositories - if there is no `builtin-spack-packages-ref`, `spack-config` informs it.

## The PR

* Update default of `builtin-spack-packages-ref` to `repos.yaml`, or `develop` if not set. 

## Testing 

Tested `yq` locally in the cases for:
* No ref of any kind
* `branch`/`tag`/`commit`s in `repos.yaml`
* Set via `builtin-spack-packages-ref`


harshula (approved):
🤞

PR #274, opened 2026-02-18
containers/Dockerfile: return the dev target and add support to build v0.22

harshula:
During the Spack v0.22 to v1.1 migration, we have noticed some supported software is not building with Spack v1.1. We may need to support Spack v0.22 CI, CD and debugging for software that needs to be released in the next few weeks.

### DIFF
--- a/containers/Dockerfile
+++ b/containers/Dockerfile
@@ -107,8 +107,8 @@ RUN zypper ref \
     vim \
  # Install gh
  && zypper addrepo https://cli.github.com/packages/rpm/gh-cli.repo \
- && zypper ref \
- && zypper install gh \
+ && zypper --gpg-auto-import-keys refresh \
+ && zypper -n install gh \
  # zypper cleanup
  &&  zypper clean
 
@@ -125,6 +125,7 @@ ARG SPACK_CONFIG_DIR=/opt/spack-config/${SPACK_VERSION}/ci
 ARG SOURCE_COMPILERS_SPACK_MANIFEST=upstream/prod/compilers.spack.yaml
 ARG SOURCE_PACKAGES_SPACK_MANIFEST=upstream/prod/packages.spack.yaml
 
+ENV SPACK_VERSION=${SPACK_VERSION}
 ENV SPACK_ROOT=/opt/spack
 ENV GNUPGHOME=${SPACK_ROOT}/opt/spack/gpg
 ENV ENV_SPACK_TARGET=${SPACK_TARGET}
@@ -150,7 +151,7 @@ SHELL ["/bin/bash", "-c"]
 
 # Install spack
 RUN <<EOF
-git clone -c feature.manyFiles=true ${SPACK_GIT_URL} ${SPACK_ROOT} --branch ${SPACK_REPO_VERSION}
+git clone ${SPACK_GIT_URL} ${SPACK_ROOT} --branch ${SPACK_REPO_VERSION}
 
 # Install ACCESS-NRIs spack-config repo
 git clone  https://github.com/ACCESS-NRI/spack-config.git /opt/spack-config --branch ${SPACK_CONFIG_REPO_VERSION}
@@ -176,7 +177,7 @@ SHELL ["docker-shell"]
 
 # https://spack.readthedocs.io/en/latest/bootstrapping.html
 # Spack is configured to bootstrap its dependencies and repositories lazily by default;
-# i.e. the first time they are needed and can’t be found.
+# i.e. the first time they are needed and can't be found.
 
 # Bootstrap spack and clone relevant spack-packages repos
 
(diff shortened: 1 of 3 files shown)


CodeGat (commented):
Although I understand the intention, the `build-ci` `v2`/`v3` split was for `0.22` vs `1.X`, so we didn't have to deal with an ever-expanding list of case statements for new things introduced between versions. . 
[`build-ci v2`](https://github.com/ACCESS-NRI/build-ci/tree/v2) still works and is used, so I would build the image from that branch instead.

harshula:
### Testing
```
$ docker run -it --rm spack-rocky:v0.22.5-20260219 

[root@460352b7abf9 /]# spack -V
0.22.5 (422a81b3d75f3f119e7c06234418763803a4f070)

[root@460352b7abf9 /]# git -C opt/spack log
commit 422a81b3d75f3f119e7c06234418763803a4f070 (HEAD -> releases/v0.22, origin/test-v0.22, origin/releases/v0.22)
Author: Matt Thompson <matthew.thompson@nasa.gov>
Date:   Wed Apr 23 09:53:00 2025 -0400

    esmf: add v8.8.1 (#28)
    
    Cherry-picked from upstream commit https://github.com/spack/spack-packages/commit/7e14786ccab85f27e717c4bd804410c5dac23847
```

harshula:
Hi @CodeGat , There is NO "ever-expanding list of case statements". There are significant difference before Spack v1 and after Spack v1. There's no point trying to maintain the same file across two branches when maintaining it on the latest branch meets our requirements.

CodeGat:
No list of case statements _yet_ - but it's not a pattern I want to start, even for the innocuous change here. 

It was a deliberate choice to split build-ci versions cleanly across major spack version lines, and I prefer to keep it that way. The compose files are set up for an explicitly spack-v1-style build of images, and likewise for v0.22 on the v2 branch.

harshula:
Hi @CodeGat , I need a Dockerfile for spack-dev images with more flexibility than that. If you're unwilling to add the flexibility to build a spack-dev image with any version of Spack, I'll have to create it elsewhere. The Dockerfile having extra options that CI doesn't need is not a problem for the compose files. e.g. The Dockerfile already has an opensuse base image option that CI never uses.

harshula:
Hi @CodeGat , Can you please check if you can build a CI container image from the current Dockerfile?

CodeGat:
Hmm, there are currently issues, it seems. But it was successfully built at d72c422de280fd08bbcc66fc3fa0200e521c3cc5, so I'll need to determine if a `spack-config` change or `packages.spack.yaml` change has brought this on...

```
#22 1.076 Installing common packages
#22 2.252 ==> Compilers have been configured automatically from PATH inspection
#22 2.285 ==> Starting concretization pool with 6 processes
#22 31.99 ==> Warning: using "cmake@3.31.6" which is a deprecated version
#22 34.03 ==> Warning: using "cmake@3.31.6" which is a deprecated version
#22 43.72 ==> Warning: using "cmake@3.31.6" which is a deprecated version
#22 47.68 ==> Warning: using "cmake@3.31.6" which is a deprecated version
#22 81.19 ==> Error: failed to concretize `netcdf-fortran@4.6.1 target=x86_64 %gcc@13.2.0 ^openmpi@5.0.3` for the following reasons:
#22 81.19      1. Only external, or concrete, compilers are allowed for the fortran language
#22 81.19      2. Only external, or concrete, compilers are allowed for the cxx language
#22 81.19      3. Only external, or concrete, compilers are allowed for the c language
```

Will look into this on Tuesday, when I'm back :)

harshula:
Hi @CodeGat , It's because of https://github.com/spack/spack-packages/issues/3505 . And fixed in `build-ci` by PR https://github.com/ACCESS-NRI/build-ci/pull/277 .

harshula:
Hi @CodeGat , Now that the Dockerfile has been cleaned up, this PR is going to return the `dev` target. The Spack v0.22 support will only be in the `dev` target and will not impact the CI related targets. Can you please review the changes?

containers/Dockerfile
```diff
@@ -188,6 +189,41 @@ SHELL ["docker-shell"]
 #   spack repo update access.nri --commit $access_spack_packages_sha
 RUN spack bootstrap now
 
+################################################################################
+FROM base-spack AS dev
+
+RUN <<EOF
+spack install gcc@13.2.0 ~binutils target=x86_64
+spack install intel-oneapi-compilers-classic@2021.10.0 target=x86_64
+spack install intel-oneapi-compilers@2025.2.0 target=x86_64
```
CodeGat: Do you have any need for the `containers/upstream/dev` directory @harshula ? I'd initially created it for the case that you were going to use the `upstream` target image as your development image.  Is there some way we can work the `upstream` target into your development workflow?

CodeGat (approved):
After a chat offline, this looks good!

PR #276, opened 2026-02-20
Remove tmate install from base-os-rocky stage

### LINKED ISSUE #275: Dockerfile: clean-up duplicate code
There appears to be duplicate code related to `tmate` and yq that should be cleaned-up.

CodeGat:
Closes #275

## Background

`tmate` is installed in the `base-os-rocky` stage, but is also installed in the `runner` and `upstream` stages. We don't need to install it in the base stage. 

## The PR

* Remove the installation from the `base-os-rocky` stage
* Move a label from the `upstream-base` to the `upstream` stage


harshula (approved):
🎉

PR #281, opened 2026-02-26
Image Building CI: Attach Volume, Use `ssh` Directly

### LINKED ISSUE #278: Image Building CI: Exit Unsuccessfully If Image Build Fails
See run https://github.com/ACCESS-NRI/build-ci/actions/runs/22328928616/job/64606523257#step:7:2669

## Background

Rather than the job failing, it continues successfully in the case the image fails to build, which is confusing.

### LINKED ISSUE #280: Image Building CI: Attach Ephemeral Volume To Docker Instance
Currently, the base Nectar VM runs out of space most runs. 

We need to attach ephemeral volumes to the instance as needed, and detach them at the end.

CodeGat:
Closes #280
Closes #278

## Background

There are a few small bugs that we can fix.

Firstly, the VMs are too small to build some images. To fix this, we create a larger bootable volume and attach the VM to that.
Secondly, because `openstack server ssh` is a thin wrapper around `ssh`, we don't get the exit status if the remote commands fail - only if the connection itself does. So we just leap over the middleman and use `ssh` directly, instead. 

## The PR

- **Boot server from large ephemeral volume**
- **Replace `openstack server ssh` with just `ssh` so we get exit status**

## Testing

Success, see https://github.com/ACCESS-NRI/build-ci/actions/runs/22468761821/job/65080775359

### DIFF
--- a/.github/workflows/containers-ci.yml
+++ b/.github/workflows/containers-ci.yml
@@ -82,6 +82,7 @@ jobs:
         # Create and assign a floating IP as part of the private runner-buildbox
         # network so GitHub Actions can SSH into it.
         run: |
+          echo "Creating boot volume..."
           openstack volume create ${{ env.NECTAR_VOLUME_NAME }} \
             --size 40 \
             --image "NeCTAR Ubuntu 22.04 LTS (Jammy) amd64 (with Docker)" \
@@ -91,6 +92,7 @@ jobs:
             &> /dev/null
           echo "Created boot volume."
 
+          echo "Creating instance..."
           openstack server create ${{ env.NECTAR_INSTANCE_NAME }} \
             --flavor p3.xxlarge \
             --key-name runner-buildbox \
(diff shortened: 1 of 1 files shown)


.github/workflows/containers-ci.yml
```diff
@@ -77,38 +78,64 @@ jobs:
...
         id: create
-        # Create ephemeral, powerful Nectar Instance, create and assign a floating IP as part of
-        # the private runner-buildbox network so GitHub Actions can SSH into it.
+        # Create ephemeral, powerful Nectar Instance, booted from a large volume.
+        # Create and assign a floating IP as part of the private runner-buildbox
+        # network so GitHub Actions can SSH into it.
         run: |
+          echo "Creating boot volume..."
+          openstack volume create ${{ env.NECTAR_VOLUME_NAME }} \
+            --size 40 \
```
aidanheerdegen, line 87: Is the size in GB?

aidanheerdegen (approved):
LGTM

.github/workflows/containers-ci.yml
CodeGat: Yep!

PR #282, opened 2026-03-01
Update Image to `2026.02.003` (Includes `access-spack-packages` directory change)

CodeGat:
## Background

See https://github.com/ACCESS-NRI/spack-config/pull/103 and https://github.com/ACCESS-NRI/access-spack-packages/pull/389

We are moving `access-spack-packages` from a spack-managed location back to a place users expect: `$spack/../access-spack-packages`. This PR rebuilds the image with the directory in the regular place, so we don't have to re-copy the directory once `spack-config` is updated. 

## The PR

* Update image to `rocky-v1.1-2026.02.003`, post-`spack-config` update that updated the location of `access-spack-packages`.

### DIFF
--- a/containers/compose.prod.yaml
+++ b/containers/compose.prod.yaml
@@ -9,7 +9,7 @@ x-common-args: &common-args
 services:
   upstream:
     container_name: Upstream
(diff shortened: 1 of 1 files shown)


harshula (approved):
🤞

PR #290, opened 2026-03-27
GitHub Spack Cache Eviction Strategy

### LINKED ISSUE #288: Cache Eviction Strategy For Buildcaches
## Background

In both self-hosed and GitHub-hosted cases, `build-ci` bumps its head on cache sizing limitations. Specifically:

* Self hosted runners have a current maximum buildcache size of 50GB in the kubernetes cluster. This can be increased somewhat, but we also have volume limitations imposed by Nectar. 
* For both self-hosted and especially GitHub-hosted runners, the OCI-backed GitHub buildcache seemingly(?) doesn't have a limit, but I'm sure it will come up eventually. In any case, inspecting docker manifests for a particular package will get longer as more packages are created. And, in the future, we will push up root specs for the purposes of a separate testing job (not just the dependencies) - these can be deleted after a while. 

We need to come up with a cache eviction strategy for each of these cases, and denote how long before something becomes evicted. 

### k8s Buildcaches

* Can be a scheduled job to create a volume debugger pod, log in to the volume, and find and delete packages from the buildcache directly. Need to determine if there are any lists of package hashes in the buildcache we need to get rid of too. 
* Could also be a `build-ci-runner` pod that sets up the mirror and `spack buildcache prune --keeplist FILE` in which the `FILE` is populated with package hashes created within the time period. 

### OCI buildcaches

* Could be a scheduled job that checks the "versions" (read: buildcache entries) of a buildcache (see https://docs.github.com/en/rest/packages/packages?apiVersion=2026-03-10#list-package-versions-for-a-package-owned-by-an-organization), and deleting the version if it is in the eviction period (see https://docs.github.com/en/rest/packages/packages?apiVersion=2026-03-10#delete-package-version-for-an-organization)

CodeGat:
Closes #288

## Background

Currently, every package is pushed to the GitHub Spack Cache, and none are deleted. Eventually, we will hit our head on some limit from GitHub, so we should proactively delete entries that are stale. 

Unfortunately, we don't have any way to know when an image version (AKA a package) was last pulled, so the next best thing we have is to evict based on last modified time. 

## The PR

- **Create github-cache-eviction.yml reusable, for removing entries from a GitHub buildcache**
- **Also create scheduled-cache-eviction.yml that calls github-cache-eviction.yml**
- **Update README.md, reworked headings**

## Testing

Tested commands in sequence:

```sh
eviction_time=$(date --universal --date "-90 days" +%FT%TZ)  # 2025-12-27T00:49:21Z
gh api --paginate --slurp /orgs/access-nri/packages/container/build-ci-buildcache/versions > versions.json
cat versions.json  # ALL versions defined
# [[{"id":751523677,"name":"sha256:679ff5dff717cace63e5dcbc00d1ce22061eaba30ae6ced96230b0e6fa2a4a6c","url":"https://api.github.com/orgs/ACCESS-NRI/packages/container/build-ci-buildcache/versions/751523677","package_html_url":"https://github.com/orgs/ACCESS-NRI/packages/container/package/build-ci-buildcache","created_at":"2026-03-23T03:41:39Z","updated_at":"2026-03-23T03:41:39Z","description":"netcdf-fortran@=4.6.1~doc+pic+shared build_system=autotools platform=linux os=rocky8 target=x86_64","html_url":"https://github.com/orgs/ACCESS-NRI/packages/container/build-ci-buildcache/751523677","metadata":{"package_type":"container","container":{"tags":["netcdf-fortran-4.6.1-dbzypnglwn4bq6vwr2kikfpdcxsz5q45.spack"]}}}, ...]]
jq -cr --arg e "$eviction_time" '[.[][] | select(.updated_at < $e)]' versions.json > evictions.json
cat evictions.json  # All versions older than 90 days
# [{"id":611140504,"name":"sha256:c15af82553ae55c502e2953a9606542407a8a7ea14ee785fd469f5ccf67807ca","url":"https://api.github.com/orgs/ACCESS-NRI/packages/container/build-ci-buildcache/versions/611140504","package_html_url":"https://github.com/orgs/ACCESS-NRI/packages/container/package/build-ci-buildcache","created_at":"2025-12-15T03:37:13Z","updated_at":"2025-12-15T03:37:13Z","description":"netcdf-fortran@=4.5.2~doc+pic+shared build_system=autotools patches:=b050dbd platform=linux os=rocky8 target=x86_64","html_url":"https://github.com/orgs/ACCESS-NRI/packages/container/build-ci-buildcache/611140504","metadata":{"package_type":"container","container":{"tags":["netcdf-fortran-4.5.2-hqifenmxf2ggq4kqk26jtwbdqzxvu657.spack"]}}}, ...]
eviction_ids=$(jq -cr '[.[] | .id]' evictions.json)  # [751523677,611140504, ...]
# Testing just one...
gh api --method DELETE /orgs/access-nri/packages/container/build-ci-buildcache/versions/751523677
gh api /orgs/access-nri/packages/container/build-ci-buildcache/versions/751523677
# No longer exists!
```

### DIFF
--- a/.github/workflows/github-cache-eviction.yml
+++ b/.github/workflows/github-cache-eviction.yml
@@ -12,7 +12,7 @@ on:
         required: true
         description: |
           Name of the package that is used as an OCI-backed spack buildcache
-      eviction-threshold:
+      eviction-threshold-days:
         type: number
         required: true
         description: |
@@ -30,11 +30,11 @@ jobs:
     env:
       GH_TOKEN: ${{ secrets.token }}
     steps:
-      - name: Get all entries in buildcache older than ${{ inputs.eviction-threshold }} days
+      - name: Get all entries in buildcache older than ${{ inputs.eviction-threshold-days }} days
         id: entries
         # The date format is ISO 8601 date/time string with a literal Z for UTC - as GitHub uses. Eg. 2025-08-07T01:50:59Z
         run: |
-          eviction_time=$(date --universal --date "-${{ inputs.eviction-threshold }} days" +%FT%TZ)
+          eviction_time=$(date --universal --date "-${{ inputs.eviction-threshold-days }} days" +%FT%TZ)
           echo "::notice::Evicting entries older than ${eviction_time}"
 
           # See https://docs.github.com/en/rest/packages/packages?apiVersion=2026-03-10#list-package-versions-for-a-package-owned-by-an-organization
@@ -58,7 +58,7 @@ jobs:
 
           echo "eviction-ids=$eviction_ids" >> $GITHUB_OUTPUT
 
-      - name: Evict entries in buildcache older than ${{ inputs.eviction-threshold }} days
+      - name: Evict entries in buildcache older than ${{ inputs.eviction-threshold-days }} days
         run: |
           for entry in ${{ steps.entries.outputs.eviction-ids }}; do
             echo "Evicting $entry..."
--- a/.github/workflows/github-cache-eviction.yml
+++ b/.github/workflows/github-cache-eviction.yml
@@ -1,0 +1,69 @@
+name: Github Cache Eviction
+on:
+  workflow_call:
+    inputs:
+      buildcache-package-org:
+        type: string
+        required: true
+        description: |
+          Organisation that owns the package that is used as an OCI-backed spack buildcache
+      buildcache-package-name:
+        type: string
+        required: true
+        description: |
+          Name of the package that is used as an OCI-backed spack buildcache
+      eviction-threshold:
+        type: number
+        required: true
+        description: |
+          Number of days after which a buildcache entry is considered stale and can be evicted
+    secrets:
+      token:
+        required: true
+        description: |
+          GitHub token with permissions to delete packages in the registry, typically a PAT with `delete:packages` scope.
+          This is needed to delete old buildcache entries from the OCI registry.
+jobs:
+  eviction:
+    name: Evict Old Entries
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.token }}
+    steps:
+      - name: Get all entries in buildcache older than ${{ inputs.eviction-threshold }} days
+        id: entries
+        # The date format is ISO 8601 date/time string with a literal Z for UTC - as GitHub uses. Eg. 2025-08-07T01:50:59Z
+        run: |
+          eviction_time=$(date --universal --date "-${{ inputs.eviction-threshold }} days" +%FT%TZ)
+          echo "::notice::Evicting entries older than ${eviction_time}"
+
+          # See https://docs.github.com/en/rest/packages/packages?apiVersion=2026-03-10#list-package-versions-for-a-package-owned-by-an-organization
+          gh api --paginate --slurp \
+            /orgs/${{ inputs.buildcache-package-org }}/packages/container/${{ inputs.buildcache-package-name }}/versions
+            > versions.json
+
+          jq --compact-output --raw-output --arg e "$eviction_time" \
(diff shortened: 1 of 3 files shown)


.github/workflows/github-cache-eviction.yml
```diff
@@ -0,0 +1,69 @@
...
+    runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ secrets.token }}
+    steps:
+      - name: Get all entries in buildcache older than ${{ inputs.eviction-threshold-days }} days
+        id: entries
+        # The date format is ISO 8601 date/time string with a literal Z for UTC - as GitHub uses. Eg. 2025-08-07T01:50:59Z
+        run: |
+          eviction_time=$(date --universal --date "-${{ inputs.eviction-threshold-days }} days" +%FT%TZ)
+          echo "::notice::Evicting entries older than ${eviction_time}"
```
atteggiani: Should we be extra specific with something like:
```suggestion
          echo "::notice::Evicting entries last-updated before ${eviction_time}"
```
?

.github/workflows/github-cache-eviction.yml
```diff
@@ -0,0 +1,69 @@
...
+
+          jq --compact-output --raw-output --arg e "$eviction_time" \
+            '[.[][] | select(.updated_at < $e)]' \
+            versions.json
+            > evictions.json
+
+          eviction_ids=$(jq --compact-output --raw-output \
+            '[.[] | .id] | join(" ")'
+            evictions.json
+          )
```
atteggiani, line 53: In case `evictions.json` is empty (all package versions have been last-updated after the eviction time) I think this step would fail. 
I'm not sure if this case could actually present itself, but might be useful to guard against that (maybe exit if the json file is empty?).

.github/workflows/scheduled-cache-eviction.yml
```diff
@@ -0,0 +1,15 @@
...
+    - cron: 30 2 3 * *
+jobs:
+  eviction:
+    permissions:
+      packages: write
+    uses: ./.github/workflows/github-cache-eviction.yml
+    with:
+      buildcache-package-org: ${{ github.repository_owner }}
+      buildcache-package-name: build-ci-buildcache
+      eviction-threshold-days: 90
```
atteggiani, line 13: This means that if all package versions in the cache have been updated within 90 days, nothing will get deleted.
I'm not sure if there could ever be the case in which we reached a GitHub limit so no more packages versions can be added, but all packages in the cache have been updated within 90 days. So this scheduled workflow wouldn't be able to delete anything.
We could always run the action manually with a lower number of `eviction-threshold-days`, but I was just wondering if `90` was chosen based on anything specific. 

Noting that I am not too familiar with the whole infrastructure, so I might be posing a case that would never exist in real-world scenarios.

atteggiani (commented):
Looks good!

I left 3 minor comments. 
2 of those are mainly coming from my lack of knowledge of the whole infrastructure, and might not really make sense in actual production. Please just disregard those if that's the case! 😃

.github/workflows/github-cache-eviction.yml
CodeGat: I think this case is fine - nulls in the filter simply aren't evaluated:

```sh
jq -n '[]' test.json
jq '[.[] | .id] | join(" ")' test.json
# ""
```

.github/workflows/scheduled-cache-eviction.yml
CodeGat: I picked 90 based on a comment by @aidanheerdegen, noting that (unless we find out we have a limitation) we won't really have to evict very much stuff. This can always be changed if we need to be more aggressive in the future.

atteggiani (approved):
Thanks @CodeGat, looks good!

PR #293, opened 2026-04-01
GitHub-Hosted Pull Fix, Spack Config Reorganisation

CodeGat:
> This is a combination of both a GitHub-hosted pull fix and the spack-config reorganisation


CodeGat:
Lets not combine these two features...

PR #292, opened 2026-04-01
`spack-config` Reorganisation

CodeGat:
References https://github.com/ACCESS-NRI/spack-config/issues/41
Closes https://github.com/ACCESS-NRI/build-ci/issues/286

## Background

We are reorganising the `spack-config` structure to make use of `includes.yaml` and removing symlinks from that file structure. See https://github.com/ACCESS-NRI/spack-config/pull/110.

## The PR

* In the Dockerfile:
  * We no longer need references to `SPACK_CONFIG_DIR` - as the config used is informed by the `include.yaml`
  * Removed all symlinking from the Dockerfile
  * Updated all build-time buildcaches to `user` (root) scope
  * Pushed down the bootstrapping of spack to `upstream`/`runner`/`dev`, allowing us to set the `ACCESS_SPACK_ADMIN` environment variable in `upstream` pre-bootstrap. 
  * Remove the runner set buildcache setup from the Dockerfile, as it was going to be set in the `user` (root) scope rather than the `user` (github) scope. It is now set up in the CI :point_down: 

* In the CI workflows:
  * Removed a bunch of convoluted logic related to re-symlinking
  * Update the upstream volume mount location from `/opt/upstream` to `/opt/release`, as the runner install tree is now `/opt/runner/release`.
  * Add setup of runner set buildcache as it was removed from the Dockerfile

## Testing

Tested in https://github.com/ACCESS-NRI/access-test-component/actions/runs/24226636073 in it's own cluster

PR #295, opened 2026-04-14
Enable spack before adding runner set buildcache back

CodeGat:
See https://github.com/ACCESS-NRI/access-spack-packages/actions/runs/24382950802/job/71210288439

## Background

A regression introduced in https://github.com/ACCESS-NRI/build-ci/pull/292, in which I didn't enable spack before running the command.

## The PR

* Enable spack before running it to create the runner set mirror


harshula (approved):
🤞

PR #297, opened 2026-04-15
GitHub-Hosted Image Fix

### LINKED ISSUE #268: GitHub-Hosted: Fails to Pull Image, Larger than Runner Disk Space
## Background

GitHub Runners only guarantee 14GB of usable space on GitHub-hosted runners, which is smaller than the `:rocky-v1.1-2025.12.001` image (17GB!) - meaning the image is too large to run on GitHub-hosted runners. Previous images were just under that size limit (eg. `:rocky-v1.1-2025.12.000` was 14GB).

## Implementation Details

Since the whole point of the GitHub host was to avoid spinning up Kubernetes runners for each manifest, we need to work within this limit. 

* See if we can rely on the OCI-backed buildcache to pull the nessecary compiler and package binaries. 

Alternatively (but not recommended):

* Create a separate image under the Runner Limit

CodeGat:
Closes #268

## Background

The upstream image that we used to run solely for GitHub-hosted invocations of `build-ci` is too large for GitHub-hosted runners to cope with. 

Rather than slimming down the image, We will instead rely solely on a binary buildcache for the install. 

### Caveats

This is slower than self-hosted runners, and slower than using an `upstream` image as a base, as not having an upstream means the concretizer will not reuse upstream-installed packages, which means that cache misses are more likely. 

Some alternatives to do if MCRs are consistently missing the cache:

* Explicitly set `inputs.spack-ref`/`inputs.spack-config-ref`/`inputs.builtin-spack-packages-ref`/`inputs.access-spack-packages-ref` to a commit SHA. This means that the concretization decisions should not vary based on different repo versions - and therefore we will almost always have cache hits. 
* In the case of compilers defined in a manifest as part of `inputs.spack-compiler-manifest-path`, use the long hash of a compiler pushed to the buildcache for perfect reuse every time:
  ```yaml
  spack:
    specs:
    - /aorj2398u2rio2kr209r209ir20
    # ...
  ```
* The above can work for packages also, in the manifest defined in the `inputs.spack-manifest-path`. 

## The PR

* Run only the `build-ci-runner` image, regardless of `inputs.run-self-hosted`, as we can't use the `build-ci-upstream` as a base. 
* Disable the spack upstream if not `inputs.run-self-hosted`
* Push Compilers installed locally to a buildcache
* Update `--scope=site` to `--scope=user` in some places

## Testing

Tested in https://github.com/CABLE-LSM/CABLE/pull/654 - it takes slightly longer to reinstall compilers from the cache, but this is unavoidable. Furthermore, since there is no upstream that can be used to influence the concretization, there is more of a chance that the hashes will be different from what is in the buildcache. 

Example of both a perfect hit on the buildcache [for compiler and manifest](https://github.com/CABLE-LSM/CABLE/actions/runs/24433564801/job/71382850677#step:22:31), and [just the compiler](https://github.com/CABLE-LSM/CABLE/actions/runs/24433564801/job/71382850675#step:22:27). This should be alleviated somewhat while the buildcache fills up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -346,7 +346,7 @@ jobs:
         if: inputs.spack-oci-buildcache-url != ''
         run: |
           . ${{ steps.env.outputs.SPACK_ROOT }}/share/spack/setup-env.sh
-          spack mirror add --scope=site \
+          spack mirror add --scope=user \
             --unsigned \
             --oci-username-variable OCI_USERNAME \
             --oci-password-variable OCI_PASSWORD \
@@ -384,7 +384,7 @@ jobs:
           spack env deactivate
           yq '.spack.specs[]' ${{ inputs.spack-compiler-manifest-path }} | while read compiler; do
             spack load $compiler
-            spack compiler find --scope=site
+            spack compiler find --scope=user
           done
           spack compiler list
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -372,6 +372,7 @@ jobs:
       - name: Manifest - Install Compilers Locally
         # Some compilers may not be available upstream, so we install them locally in a separate spack.yaml
         if: inputs.spack-compiler-manifest-path != ''
+        id: compilers
         env:
           # For pulling OCI buildcache
           OCI_USERNAME: ${{ secrets.spack-oci-buildcache-username || github.actor }}
@@ -387,6 +388,17 @@ jobs:
             spack compiler find --scope=user
           done
           spack compiler list
+
+      - name: Manifest - Push Local Compilers to Buildcache
+        env:
+          # For pulling and pushing OCI buildcache
+          # If secrets.GITHUB_TOKEN has permissions.packages:write, one would not have to specify secrets.spack-oci-buildcache-password
+          OCI_USERNAME: ${{ secrets.spack-oci-buildcache-username || github.actor }}
+          OCI_PASSWORD: ${{ secrets.spack-oci-buildcache-password || secrets.GITHUB_TOKEN }}
+        if: inputs.spack-oci-buildcache-url != '' && steps.compilers.outcome == 'success'
+        run: |
+          . ${{ steps.env.outputs.SPACK_ROOT }}/share/spack/setup-env.sh
+          spack -e local_compilers buildcache push --unsigned oci_buildcache
 
       - name: Manifest - Install jinja-cli
         run: python3 -m pip install jinja-cli==1.2.2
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -266,6 +266,19 @@ jobs:
...
 
+      - name: Init - Disable Self-Hosted Upstreams
+        if: ${{ ! inputs.run-self-hosted }}
+        # For GitHub-hosted versions of build-ci, we don't have an upstream, so we disable in our user scope
+        # NOTE: In future may be able to be replaced by spack config --scope=user add upstreams::{}
+        run: |
+          . ${{ steps.env.outputs.SPACK_ROOT }}/share/spack/setup-env.sh
+
+          echo "--- Disable upstream ---"
+          upstream_location=$(spack config --scope=user edit upstreams --print-file)
```
aidanheerdegen, line 277: Test is [saying upstream not found](https://github.com/CABLE-LSM/CABLE/actions/runs/24433564801/job/71382850677#step:5:13)

```
==> Warning: upstream not found: /opt/release/.spack-db/index.json
```

.github/workflows/ci.yml
```diff
@@ -372,21 +385,34 @@ jobs:
...
+      - name: Manifest - Push Local Compilers to Buildcache
+        env:
+          # For pulling and pushing OCI buildcache
+          # If secrets.GITHUB_TOKEN has permissions.packages:write, one would not have to specify secrets.spack-oci-buildcache-password
+          OCI_USERNAME: ${{ secrets.spack-oci-buildcache-username || github.actor }}
+          OCI_PASSWORD: ${{ secrets.spack-oci-buildcache-password || secrets.GITHUB_TOKEN }}
+        if: inputs.spack-oci-buildcache-url != '' && steps.compilers.outcome == 'success'
+        run: |
+          . ${{ steps.env.outputs.SPACK_ROOT }}/share/spack/setup-env.sh
+          spack -e local_compilers buildcache push --unsigned --private oci_buildcache
```
aidanheerdegen, line 414: Are compilers not automatically pushed to a build cache?

CodeGat: Nothing is pushed to the OCI buildcache automatically, just to split install time from package push time.

.github/workflows/ci.yml
CodeGat: Yep! and that will be the only time it says it - when we are running `$spack/share/spack/setup-env.sh` as part of that step, before disabling it. Later steps won't have that warning.

CodeGat: An alternative is to start with no upstream in the `spack-config`, and enable it in the workflow instead :thinking:

aidanheerdegen (approved):

Issue #298, opened 2026-04-15
Where Should Tests Be Stored?


CodeGat:
See related comment (and above):

> Some of these decisions are better informed by discussing with the model development teams about how they would like to structure their tests.
> 
> If they want to keep the tests separate from the codebase, then `.github` or even a separate repo might be attractive.
> 
> If they want the exact opposite, and would like the tests sitting in with the code, and runnable by any user who clones the repo just by invoking `pytest`, then they might want the opposite.
> 
> We may want to support both, or maybe 90% want one approach, so we should prioritise that in the first instance, while keeping support for other approaches alive as possibilities, but maybe not supported in the first iteration.
> 
> This is my way of not really giving an answer at all. 

 _Originally posted by @aidanheerdegen in [#289](https://github.com/ACCESS-NRI/build-ci/issues/289#issuecomment-4204372787)_

## Background

From least to most customisable:

* Should we let `pytest` discover all pytest files in the MCR, regardless of directory?
* Should we assume tests are stored under `.github/build-ci/tests` of the MCR?
* Should it be a default to a potential `inputs.pytest-test-root`?
* Should we not even assume that tests will be stored in the MCR, and instead have a `inputs.pytest-test-repository[-ref]`?

This will require meeting with MCR maintainers to see what they're hoping to do.

CodeGat:
Currently the prototype is using option 3 - a `inputs.pytest-search-path` that defaults to `.` but would usually be `.github/build-ci/tests`, given pytest does not by default search hidden directories.

PR #299, opened 2026-04-15
Post-Build Testing

### LINKED ISSUE #193: Consider Post-Build Testing Requirements

CodeGat:
Closes #193 

## Background

Right now, we install the component manifests but don't do any testing outside of the installation status of `spack`. We need a framework for post-build testing, and right now we're going with `pytest` due to the strength of test specification via `--marker`s, it's use as a testing meta-framework in other repositories such as `ACCESS-NRI/model-config-tests`, and it's integration with GitHub Actions. 

Essentially we will have a second job (`Test`) after the first (`Install`), that will reinstall the lockfile via a buildcache, then discover all tests in the MCR, and run them. Users can specify args to `pytest` at the manifest level via a reserved definition `_pytest-args`, like so:

```yaml
spack:
  definitions:
  - _pytest-args: ['-m marker1 and not marker2', '--verbosity=1']
  # ... and the rest
```

> [!NOTE]
> The exact machinations of pytest discovery will be based on the resolution of https://github.com/ACCESS-NRI/build-ci/issues/298

## The PR

* Add new `test-build` dependent job in the `ci.yml` workflow that install the previously installed-and-pushed packages from the lockfile, then runs `pytest`
* Optional args to `pytest` are given at the manifest level via a `_pytest-args` reserved definition in the manifest. 
* Consolidated a large amount of setup code for both jobs under the new `.github/actions/setup-instance` action.

### Proof of Concept

Tested in https://github.com/ACCESS-NRI/access-test-component/pull/19, specifically https://github.com/ACCESS-NRI/access-test-component/actions/runs/24760630710?pr=19. Note the tests under `.github/build-ci/tests`, discovered by `inputs.pytest-search-path` (as hidden directories aren't found traditionally. 

Also note the reading of reserved definitions, for example written in https://github.com/ACCESS-NRI/access-test-component/pull/19/changes#diff-b5e432583957cd6026f987a7ddf4b2df3a0daab1e7afd40293562f16c0d6b4b1R3 and added successfully in https://github.com/ACCESS-NRI/access-test-component/actions/runs/24760630710/job/72443635335#step:11:8.

Installation in the test job is _much_ quicker when using a lockfile and `spack install --use-cache` - possibly because we skip concretization and go to querying the buildcache immediately - from 2m with perfect reuse [in the initial install job](https://github.com/ACCESS-NRI/access-test-component/actions/runs/24432348048/job/71379298965?pr=19#step:21:27) to 20s with perfect reuse [in the later test job](https://github.com/ACCESS-NRI/access-test-component/actions/runs/24432348048/job/71379647707?pr=19#step:15:14).

### DIFF
--- a/.github/actions/setup-instance/README.md
+++ b/.github/actions/setup-instance/README.md
@@ -1,0 +1,51 @@
+# Setup Container Instance
+
+Action that prepares a build-ci runner container instance for CI by exporting key environment variables, updating repositories to requested refs, and configuring Spack mirrors/upstreams where required.
+
+## Inputs
+
+| Name | Type | Description | Required | Default | Example |
+| ---- | ---- | ----------- | -------- | ------- | ------- |
+| `spack-ref` | `string` (git branch, tag or sha) | The git ref to checkout for the `spack` repository. | `true` | N/A | `"develop"` or `"v1.0.0"` or `"5a1cdc4e"` or `""` |
+| `spack-config-ref` | `string` (git branch, tag or sha) | The git ref to checkout for the `spack-config` repository. | `true` | N/A | `"main"` or `"v0.5.2"` or `"3f14e99"` or `""` |
+| `builtin-spack-packages-ref` | `string` (git branch, tag or sha) | The git ref to checkout for the builtin spack packages repository. | `true` | N/A | `"develop"` or `"release-1.2"` or `"a8b7c6d"` or `""` |
+| `access-spack-packages-ref` | `string` (git branch, tag or sha) | The git ref to checkout for the `access-spack-packages` repository. | `true` | N/A | `"main"` or `"v2026.04"` or `"9e72b1c"` or `""` |
+| `spack-oci-buildcache-url` | `string` (URL) | The OCI registry URL to add as a Spack buildcache mirror. Pass an empty string (`''`) to skip adding an OCI mirror. | `true` | N/A | `"ghcr.io/access-nri/spack-buildcache"` or `""` |
+| `run-self-hosted` | `string` (boolean) | Whether this instance is running on a self-hosted runner. Controls upstream disabling and runner-set buildcache setup. | `true` | N/A | `"true"` or `"false"` |
+
+## Outputs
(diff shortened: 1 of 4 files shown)


CodeGat:
Hmm...currently we activate the environment in the CI before we run `pytest`, meaning we don't have to have pre-test environment activation. But maybe that's a good thing that we're explicit, so tests outside of CI (but using spack) attempt to activate the environment. In any case, the fixture would look like:

```py
@pytest.fixture(autouse=True)
def spack_env(pytestconfig):
    spack_env_path: str | None = pytestconfig.getoption("spack_env_path", default=None)

    if not spack_env_path:
        # Don't use this fixture if the spack_env_path is not set
        return

    # Reference the default environment the packages were installed in, before each test
    env: spack.environment.Environment = spack.environment.Environment(
        spack_env_path
    )

    # Activate the spack environment, and collect the users environment variable modifications
    env_activate_mods = spack.environment.shell.activate(env)
    # Apply the environment variable modifications to the user
    env_activate_mods.apply_modifications()

    if not env.active:
        raise RuntimeError(
            f"Couldn't activate the default environment at {spack_env_path}"
        )

    # Return control to the test
    yield

    # Once the above completes, deactivate the environment and collect the now-reversed user environment modifications
    env_deactivate_mods = spack.environment.shell.deactivate()
    # Apply those modifications so we're now back where we started!
    env_deactivate_mods.apply_modifications()
```

PR #311, opened 2026-07-10
Use relative refs for all callee actions

### LINKED ISSUE #310: Use Relative Refs For Actions
## Background

Similar to ACCESS-NRI/build-cd#346, we should use relative refs for repositories using major-version branches, like `build-ci` and `build-cd`, so we don't haveto update all the `vX` refs throughout the repository when we get a new major branch. 

## Implementation Details

See https://github.com/ACCESS-NRI/build-cd/issues/346#issuecomment-4932271380

CodeGat:
Closes #310

## Background

We should use relative refs for all actions in `build-ci` - means we don't have to do any mass updates of versions. 

## The PR

* Checkout `build-ci` at the start of the job
* Use actions invocations relative to the above repository
* To prevent clobbering of the `build-ci` repository when cloning later repositories under `github.workspace`, give the manifest repository a custom `manifest` path and reference that path in all later steps under `working-directory`

## Testing

Done in https://github.com/ACCESS-NRI/access-test-component/pull/25, noting the [cloning of the `build-ci` actions repository at the right ref](https://github.com/ACCESS-NRI/access-test-component/actions/runs/29220813905/job/86725412418?pr=25#step:3:97), and the [use of that repo for later actions](https://github.com/ACCESS-NRI/access-test-component/actions/runs/29220813905/job/86725412418?pr=25#step:15:4)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -216,6 +216,13 @@ jobs:
       # Common directory for outputs to be added to artifact
       artifact-content-dir: /tmp/artifact-output
     steps:
+      - name: Init - Clone build-ci for custom actions
+        uses: actions/checkout@v6
+        with:
+          repository: access-nri/build-ci
+          ref: ${{ job.workflow_sha }}
+          path: build-ci
+
       - name: Init - Initalize dynamic vars
         id: init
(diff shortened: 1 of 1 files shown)


atteggiani (approved):
Thanks Tommy, this looks pretty good! 

I like the approach of using a local clone and then using relative calls for the actions. This same approach will solve some version-updating issues in other repositories!

